PR #447, opened 2015-05-05
Added buttons to copy to clipboard nuget commands

### LINKED ISSUE #423: Copy Saga Audit NuGet command to clipboard
Related to #422

Right after the `install-package ...` it would be really nice to have a Copy to Clipboard icon you could just click. Right now you can't even select the text and copy it that way.

johnsimons:
Fixes #423


johnsimons:
@distantcam please review

Issue #446, opened 2015-05-05
JSON Body view is slow and losses color formatting


cjoergensen:
Hi

The JSON body is very slow when loading some messages (> 8 sec) and is unable to do color formatting on some attributes:

![image](https://cloud.githubusercontent.com/assets/8087733/7468000/58ffdd60-f303-11e4-8177-85475b36832c.png)

Could be related to #357 

I cannot post the full data of the message here, as it contains confidential data. But I can send it to you directly if needed.

Kind regards,
Christian

johnsimons:
@cjoergensen 
Can you send it john.simons(at)particular.net

cjoergensen:
@johnsimons it's should be in your inbox

johnsimons:
Fixed as part of #449

johnsimons:
@cjoergensen is this working now in v1.2.9 ?

cjoergensen:
@johnsimons yes, it is much better.

johnsimons:
nice to hear :smile:

PR #449, opened 2015-05-06
New syntax highlighter

johnsimons:
Modified the xml and json viewer to use AvalonEdit instead.

Hopefully this will fix:
- #446


johnsimons:
@distantcam This is ready for a review.

Any other issues you think this change will fix ?

PR #356, opened 2015-05-07
Saga View window scroll is not consistent

johnsimons:
* Scrolling up and down the Saga View with the scroll option is fine (either with the mouse or with the up/down arrows

image

* However, the scroll behavior is erratic, and its size changes (see below and compare to the above)

image

### UX impressions

* In general the experience of navigating up/down with the mouse holding down the scroll control is a good enough UX
* Clicking with the mouse on the scroll to bring it to a location does not provide a consistent behavior
* Pressing the up/down (or page up/down) buttons also generates unpredictable results and is not a good UX.

### Proposed Fix

* Make the clicking on the the scroll area and pressing the up/down button behave similarly to holding down the mouse button and dragging the scroller to the requested location.


johnsimons:
@distantcam this one is ready for review

Issue #353, opened 2014-07-08
The hover label in the Endpoint Explorer only displays "Installed On: ".


cjoergensen:
The hover label in the Endpoint Explorer only displays "Installed On: ":

![image](https://cloud.githubusercontent.com/assets/8087733/3507266/8e96bd8c-067c-11e4-8038-ddb3f4f52442.png)

ServiceInsight Version: 1.1.4
ServiceControl Version: 1.1.0

![screen1](https://cloud.githubusercontent.com/assets/8087733/3507369/aebe7de2-067d-11e4-96fb-ce16120cecff.png)

cjoergensen:
Hi

This bug seems to be back:

![image](https://cloud.githubusercontent.com/assets/8087733/7114373/ee92c578-e1dc-11e4-8c55-bfb732488fcd.png)

___________

![image](https://cloud.githubusercontent.com/assets/8087733/7114380/042ba332-e1dd-11e4-80f8-ffe35fb06c76.png)

distantcam:
This bug is dependent on which version of ServiceControl you're using. (Yay!)

Should be fixed for the latest version of SC.

cjoergensen:
It is still a problem when using ServiceControl 1.5.2 and ServiceInsight 1.2.6.

distantcam:
Ugh. Ok apart from SC changing it's json you're also experiencing an issue because the host name just isn't set in SC.

cjoergensen:
Here is the output from the ServiceControl/api/endpoints:

{
	"id": "b4dcea58-eb93-1e79-aefb-41c21e835df6",
	"name": "[EndpointName]",
	"host_display_name": "ESBENDPOINT-P1",
	"monitored": true,
	"monitor_heartbeat": true,
	"license_status": "unknown",
	"heartbeat_information": {
		"last_report_at": "2015-04-23T06:37:37.9634959Z",
		"reported_status": "beating"
	},
	"is_sending_heartbeats": true
}

PR #444, opened 2015-05-11
Sequence diagram is not showing Hosts for endpoints

johnsimons:
The Sequence diagram is not showing Hosts for endpoints

image

Could be related #353


johnsimons:
@distantcam 
So the host display name is exposed using two different property names, for a api/conversations/{id} the host is exposed via "host" but for /api/endpoints the host is exposed via "host_display_name".

johnsimons:
Ready for review @distantcam

PR #453, opened 2015-05-11
NullReferenceException being thrown in saga view window

johnsimons:
When a user is looking at the saga window and show saga data is enabled
image

If the user selects a different saga message and that endpoint does not have the saga plugin installed, the loading of the saga fails but because the   show saga data is enabled an exception is thrown:

```
InnerException: System.NullReferenceException
       HResult=-2147467261
       Message=Object reference not set to an instance of an object.
       Source=ServiceInsight
       StackTrace:
            at Particular.ServiceInsight.Desktop.Saga.SagaWindowViewModel.RefreshMessageProperties() in c:\Projects\ServiceInsight\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 60
            at Particular.ServiceInsight.Desktop.Saga.SagaWindowViewModel.RefreshSaga(SagaInfo originatingSaga) in c:\Projects\ServiceInsight\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 130
            at Particular.ServiceInsight.Desktop.Saga.SagaWindowViewModel.RefreshSaga(StoredMessage message, Func`2 HasChanged) in c:\Projects\ServiceInsight\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 96
            at Particular.ServiceInsight.Desktop.Saga.SagaWindowViewModel.Handle(SelectedMessageChanged event) in c:\Projects\ServiceInsight\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 76
       InnerException: 
```

### DIFF
--- a/src/ServiceInsight/Saga/SagaWindowViewModel.cs
+++ b/src/ServiceInsight/Saga/SagaWindowViewModel.cs
@@ -31,6 +31,11 @@
 
         public void OnShowMessageDataChanged()
         {
+            if (Data == null || Data.Changes == null)
+            {
+                return;
+            }
+
             RefreshShowData();
 
             if (ShowMessageData)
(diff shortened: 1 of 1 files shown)


johnsimons:
Ready for review @distantcam

PR #455, opened 2015-05-13
Enabled in-memory caching of rest responses

johnsimons:
This PR enables the use of etags to do proper response caching.
With this change performance increases slightly (there is still work to do).

Some responses are immutable and therefore are cached permanently, this only applies to successful audited messages


johnsimons:
@distantcam please review

PR #458, opened 2015-05-15
Remove old endpoints from endpoint explorer

### LINKED ISSUE #168: When Refreshing endpoints should also consider removing of the nodes that are removed from SC
This can be tested once [deletion / purging of the messages][1] is done in SC but basically when you refresh the Endpoint explorer list, we need to also remove the endpoints that are no longer relevant (probably because messages in those Endpoints are purged).

[1]:https://github.com/Particular/ServiceControl/issues/114

johnsimons:
Remove the endpoints that are no longer relevant (probably because messages in those Endpoints have been purged).

Fixes #168


johnsimons:
@distantcam ready to be reviewed

Issue #391, opened 2014-10-02
Very large stack trace screws with header view scrolling


SimonCropp:
very minor but.... note that the scroll bar height is inconsistent as you scroll

![halfwayscroll](https://cloud.githubusercontent.com/assets/122666/4487414/d4a3fa9e-49fa-11e4-8237-409ea82b9f44.png)

![scollalittlefurther](https://cloud.githubusercontent.com/assets/122666/4487415/d856e278-49fa-11e4-95af-bee0d0406a20.png)

johnsimons:
Found the issue in devexpress - https://www.devexpress.com/Support/Center/Question/Details/Q434570

johnsimons:
So i think the solution is to remove the devexpress grid and use the one out of the box :smile:

johnsimons:
Fixed in #461

PR #461, opened 2015-05-20
Header viewer improvements

### LINKED ISSUE #442: Copy to clipboard does not respect selection
image

I was expecting to select whatever headers I'm interested in and then when I use the "Copy to clipboard", to only copy the highlighted headers

At the moment it copies all header regardless of the highlighted rows

johnsimons:
Fixes #442 and #391


johnsimons:
@distantcam this one is also ready to be reviewed

Majority of the change is to rip out the devexpress controls and use out of the box controls instead

PR #474, opened 2015-07-13
Scaled out instance shows as one with tooltips for all the Hosts

SeanFeldman:
This addresses issue [Endpoint explorer in SI hard to use when repeated endpoint name appear](https://github.com/Particular/FeatureDevelopment/issues/310)
Should `EndpointExplorerTests` be updated to reflect this?

@Particular/ServiceInsight please review

### DIFF
--- a/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
+++ b/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
@@ -3,14 +3,14 @@
     using System;
     using System.Linq;
     using Caliburn.Micro;
-    using ExtensionMethods;
+    using Particular.ServiceInsight.Desktop.ExtensionMethods;
     using Particular.ServiceInsight.Desktop.Framework;
     using Particular.ServiceInsight.Desktop.Framework.Events;
     using Particular.ServiceInsight.Desktop.Framework.Settings;
-    using ServiceControl;
-    using Settings;
-    using Shell;
-    using Startup;
+    using Particular.ServiceInsight.Desktop.ServiceControl;
+    using Particular.ServiceInsight.Desktop.Settings;
+    using Particular.ServiceInsight.Desktop.Shell;
+    using Particular.ServiceInsight.Desktop.Startup;
 
     public class EndpointExplorerViewModel : Screen,
         IHandle<RequestSelectingEndpoint>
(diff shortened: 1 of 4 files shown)


petersgiles:
If we have to make a lot of distinct lists is might be worth including an extension method that uses a HashSet (very fast)

    public static IEnumerable<TSource> DistinctBy<TSource, TKey>(
        this IEnumerable<TSource> source,
        Func<TSource, TKey> keySelector)
    {
        var knownKeys = new HashSet<TKey>();
        return source.Where(element => knownKeys.Add(keySelector(element)));
    }

you could call it like this 

    var distinctList = sourceList.DistinctBy(x => x);

or in this case 

    var hostNames = instances.DistinctBy(instance => instance.HostDisplayName);
    var tooltip = String.Join(",", hostNames.ToArray());

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
```diff
@@ -170,9 +170,14 @@ public void RefreshData()
...
             ServiceControlRoot.Children.Clear();
-            foreach (var endpoint in endpoints.OrderBy(e => e.Name))
+
+            var endpointInstancesGroupedByName = endpoints.GroupBy(x => x.Name);
+            foreach (var scaledOutEndpoint in endpointInstancesGroupedByName)
             {
-                ServiceControlRoot.Children.Add(new AuditEndpointExplorerItem(endpoint));
+                var instances = scaledOutEndpoint.ToList();
+                var hostNames = instances.Select(instance => instance.HostDisplayName);
+                var tooltip = string.Join(", ", hostNames);
```
johnsimons: Just wondering if it is better to stack them ? instead of comma delimited

src/ServiceInsight/SequenceDiagram/SequenceDiagramView.xaml
```diff
@@ -114,7 +114,7 @@
                                         <Label Grid.Row="0" Grid.Column="1" Content="{Binding FullName}" Style="{StaticResource TooltipLabelStyle}" />
                                         <Label Grid.Row="1" Grid.Column="0" Content="NSB Version" Style="{StaticResource TooltipLabelStyle}" />
                                         <Label Grid.Row="1" Grid.Column="1" Content="{Binding Version}" Style="{StaticResource TooltipLabelStyle}" />
-                                        <Label Grid.Row="2" Grid.Column="0" Content="Host" Style="{StaticResource TooltipLabelStyle}" />
+                                        <Label Grid.Row="2" Grid.Column="0" Content="HostNames" Style="{StaticResource TooltipLabelStyle}" />
```
johnsimons: Why this change? wouldn't in this diagram we would want to see the exact host that handled the message ?

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
SeanFeldman: Have no preference and it wasn't specified how to separate between host names. Is stacking them would be better @johnsimons ?

src/ServiceInsight/SequenceDiagram/SequenceDiagramView.xaml
SeanFeldman: Probably crippled in with R# rename. Good catch @johnsimons

johnsimons:
Thanks @SeanFeldman, when u have time can u fix the description so it looks pretty for the release notes

PR #475, opened 2015-07-16
SI throws exception when selected message ID is not a Guid

weralabaj:
NSB can handle messages coming from non-NSB endpoints, then transport level message ID is assigned. It doesn't have to be Guid (see failed message below that uses MSMQ id).
SI assumes that all IDs will be Guids, so when such a message is selected on the grid, SI throws exception and stops functioning.
capture

### DIFF
--- a/src/ServiceInsight/Saga/SagaWindowViewModel.cs
+++ b/src/ServiceInsight/Saga/SagaWindowViewModel.cs
@@ -5,6 +5,7 @@
     using System.Linq;
     using System.Windows.Input;
     using Caliburn.Micro;
+    using DevExpress.Utils.Text.Internal;
     using Models;
     using Particular.ServiceInsight.Desktop.ExtensionMethods;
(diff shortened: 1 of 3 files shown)


weralabaj:
@distantcam Please, have a look. Turns out it was only problem with saga view.

src/ServiceInsight/Saga/SagaWindowViewModel.cs
```diff
@@ -5,6 +5,7 @@
     using System.Linq;
     using System.Windows.Input;
     using Caliburn.Micro;
+    using DevExpress.Utils.Text.Internal;
```
distantcam: What's this using for?

weralabaj: Sorry, removed.

johnsimons:
Looks good @weralabaj, go ahead and merge it.
Also, what was the exception thrown ?

PR #476, opened 2015-07-16
Message Flow diagram node auto sizing and selection

johnsimons:
This issue addresses two main concerns:

- Sizing the flow diagram nodes so that message types and endpoint names are readable without having to use tooltips. This is helpful if you have long endpoint names or message types.

- Improve selection of messages directly from flow diagram. Message body and header and message properties are kept in sync with the selected node in the Flow Diagram.

There's also a few rendering and performance improvements in Message Flow diagram.


distantcam:
@johnsimons What's the feature request for this?

johnsimons:
@distantcam see updated description

distantcam:
@johnsimons That doesn't answer my question. What is the feature request for this PR?

johnsimons:
> What is the feature request for this PR?

U mean in FeatureDevelopment repo ?

distantcam:
@johnsimons Yes

johnsimons:
it is linked

weralabaj:
@johnsimons I looked into the issues we discussed this week regarding resizing. I pushed a fix for aligning nodes on the surface after resizing to [another branch](https://github.com/Particular/ServiceInsight/tree/fix-resized-nodes-alignment-on-surface).
It's a tiny change, but seems that resizing is more complicated than I first thought, do you agree that it might be better to do it separately from changing how views are refreshed?

I found some issue which I think is a bug in the control. The width recalculation is weird, sometimes it fits the widest element (usually endpoint name) perfectly (I used blue squares to show that), but sometimes there's a gap (red squares). I have no idea where that comes from and didn't find any plausible explanation regarding what it depends on.
resizing_nodes_issue

The another issue is that I added layout refresh after resizing, which fixes the problem with overlapping nodes, but then it causes "jumping" after showing/hiding endpoints. Still better than what we had before, as it's not resizing the view and I'm not sure if we can do it better easily. 
One idea might be to focus on the element that is in the middle of the screen when button is clicked and keep that element in the middle after resize. But I'm not sure yet now how much work would that be. Is it worth looking into or the current thing is "good enough"?

I'll have to test it more and make sure there are no weird edge cases, still work in progress...

johnsimons:
@weralabaj I have previously contacted the 3rd party vendor via their forums, maybe let's open another support case and see what they say?
We may need to reproduce it in isolation.

weralabaj:
@johnsimons Ok, I'll raise it again with our screens and if they need repro I'll prepare some code.

HEskandari:
Fix the border issue on b4a1f3e.

johnsimons:
@HEskandari is it ok if u refresh the original description for this PR once u done, so it reads good in release notes

PR #478, opened 2015-07-27
Added a help message to explain missing body text in message view

petersgiles:
This is part of https://github.com/Particular/FeatureDevelopment/issues/296 

I added a panel that displays some helpful information to developers who are confronted with the problem 296 explains

### DIFF
--- a/src/ServiceInsight/MessageViewers/MessageBodyView.xaml
+++ b/src/ServiceInsight/MessageViewers/MessageBodyView.xaml
@@ -12,10 +12,19 @@
              d:DesignWidth="300"
              mc:Ignorable="d">
 
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="/Saga/SagaResources.xaml" />
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
+    </UserControl.Resources>
+
     <Grid>
         <dxlc:LayoutControl Padding="0"
                             HorizontalAlignment="Stretch"
-                            VerticalAlignment="Stretch">
+                            VerticalAlignment="Stretch"
+                            Visibility="{Binding Path=NoContentHelpNotVisible, Converter={StaticResource BoolToVisibilityConverter}}">
             <dxlc:LayoutGroup View="Tabs">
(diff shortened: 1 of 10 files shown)


src/ServiceInsight/MessageViewers/MessageBodyView.xaml
```diff
@@ -28,11 +37,42 @@
...
+                <StackPanel>
+                    <TextBlock Background="{x:Null}"
+                                FontFamily="Consolas"
+                                FontSize="16"
+                                Padding="2"
+                                TextWrapping="Wrap"
+                                Style="{StaticResource CodeStyle}">
+                        <Run Text="Messages that have a large message body are not stored. " />
+                        <Run Text="This is to ensure that the majority of our users enjoy the best level of performance. " />
+                        <Run Text="For users with special analysis needs, edit MaxBodySizeToStore in ServiceControl.config to increase the size of storeable messages." />
```
johnsimons: Should it be `ServiceControl.exe.config` ?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -161,7 +161,25 @@ public void LoadBody(StoredMessage message)
...
+                var presentationBody = new PresentationBody();
+
+                switch (response.StatusCode)
+                {
+                    case HttpStatusCode.OK:
+                        presentationBody.Text = CleanupBodyString(response.Content);
+                        break;
+                    case HttpStatusCode.NoContent:
+                        presentationBody.Hint = PresentationHint.NoContent;
+                        presentationBody.Text = "Body was too large and not stored. Edit ServiceControl/MaxBodySizeToStore to be larger in the ServiceControl configuration.";
```
johnsimons: Is this text ever displayed ?

johnsimons:
looks good, added a few comments

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
petersgiles: No only put it there to assist debugging later on, if you were maintaining this a year from now and were trying to work out what the NoContent means this little snippet would be pretty useful, also would appear anywhere we just use text instead of special XAML views.

PR #479, opened 2015-07-29
Colouring stack traces in ServiceInsight

petersgiles:
As a developer reading Stack traces is a daily job, so with this new release of ServiceInsight we have spiced things up and added some colour to the boring black while Stack traces.
Hopefully this will aid you in your quest to identify and kill those nasty bugs.

Stack traces are now colored to help you pick out key data at a glance.

capture
capture2


petersgiles:
@Particular/serviceinsight  Please review

petersgiles:
This will be released in Service Insight 1.2.11

PR #500, opened 2015-09-16
Export to png

johnsimons:


johnsimons:
@weralabaj it seems the timelines are not centre aligned to the endpoint.

weralabaj:
Yes @johnsimons, this is because of margins used on one of the elements. I noticed you're making quite a lot of changes anyway so I decided to wait instead of investigating now and redoing it later.

johnsimons:
Converting existing issue to PR instead

Issue #547, opened 2015-12-23
ServiceInsight crashes when saving options on German Windows 10


gbiellem:
I was testing a release of ServiceControl on a German Windows 10 and found this issue in ServiceInsight 1.5.1
If I go to View->Options  and then change anything and press "Save"  ServiceInsight crashes completely. 
The exception is not caught by the "oops" screen. 
Tried the pre-release of 1.5.2 and same issue occurred

distantcam:
Confirmed, this is a Costura bug.

I'm thinking we don't need Costura, since SI has an installer, so I'll just remove it.

johnsimons:
I put a PR for this one

johnsimons:
This is actually a pre-step for the auto-update

gbiellem:
@johnsimons Is the auto-update still on the cards? I thought you had shelved it.

distantcam:
Fixed by #549

PR #549, opened 2016-01-04
Removing Costura

johnsimons:
Removing Costura as the language packas are not included and we don't really need it since SI is deployed via an installer


johnsimons:
@gbiellem can you please update the installer to include all files under bin/release

johnsimons:
@distantcam just to comfirm, we package all language subfolders right ? And the folder structure needs to stay intact ?
image

distantcam:
@johnsimons That's correct, the folder structure needs to remain the same.

johnsimons:
Tested installer, it work :)

@distantcam I think this is done, quick review ?

distantcam:
This fixes #547

PR #550, opened 2016-01-04
Fixing settings serialization broken by CM upgrade

distantcam:
The Caliburn Micro upgrade broke the serialization of the ProfilerSettings class. This should fix the problem.


johnsimons:
Since the  Caliburn Micro upgrade didn't really fix anything, should we just roll that back ?

mikeminutillo:
What happened to the `Math.Max` calls? Are they no longer needed?

distantcam:
Ultimately we should clean up the options screen to tell the user when a value is out of range instead of just clamping the value, but I already burnt too much time trying to fix that screen.

Issue #553, opened 2016-01-11
Desktop and start menu shortcuts are missing when upgrading to 1.5.2


icwesley:
Upgrading to 1.5.2 removes all ServiceInsight shortcuts and does not replace them.

gbiellem:
Confirmed that this is the case. It's not tied to a specific OS.

PR #554, opened 2016-01-12
Fix missing shortcuts

gbiellem:
A change in 1.5.2 broke the shortcuts.  See #553


gbiellem:
@HEskandari skandari  and @distantcam I've fixed the shortcut issue. Merge?

Issue #545, opened 2015-12-21
"Only one arrow is allowed to come in" Exception


WojcikMike:
While running pub/sub example on NSB v6 I tend to get below exception quite often. The sample code can be found in here:
https://github.com/WojcikMike/PubSubSample

While trying to open audited message (mostly while clicking EventMessage) this is the exception that I get.

System.Exception: Only one arrow is allowed to come in
   at ServiceInsight.SequenceDiagram.Diagram.Handler.set_In(Arrow value) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\SequenceDiagram\Diagram\Handler.cs:line 40
   at ServiceInsight.SequenceDiagram.ModelCreator.Initialize() in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\SequenceDiagram\ModelCreator.cs:line 106
   at ServiceInsight.SequenceDiagram.SequenceDiagramViewModel.CreateElements(List`1 messages) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\SequenceDiagram\SequenceDiagramViewModel.cs:line 227
   at ServiceInsight.SequenceDiagram.SequenceDiagramViewModel.Handle(SelectedMessageChanged message) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\SequenceDiagram\SequenceDiagramViewModel.cs:line 201

HEskandari:
I ran the sample, this looks to be a bug in ServiceControl.

When I load the conversation, I get two messages, that are pretty much identical, except for the `time_sent` and `delivery_time` (even `id` and `message_id` are the same)

![image](https://cloud.githubusercontent.com/assets/132117/11945386/8307533a-a8a3-11e5-9a22-b180a5d00145.png)

Now as per code, you are publishing an event of type `MySuperPowerAreEnabled` in your publisher, then handling it in the Subscriber2 and publish another instance of the same event type, plus a command of type `LocalCommand`.

The conversation I get from ServiceControl doesn't resemble that (missing local sent `LocalCommand`) and what it does give me (two almost identical messages) feels like a bug.

I have attached the JSON payload I get for that conversation, could a ServiceControl expert have a look? @mikeminutillo 

[Conversation.txt](https://github.com/Particular/ServiceInsight/files/69179/Conversation.txt)

Having said all that, @johnsimons reported so wierd looking diagrams when he was testing v6 stuff. It *could* be that this is a related [issue][1].

[...]

PR #555, opened 2016-01-12
Added error handler for SD with raygun reporting

distantcam:
Fix for #545 

image

The link reports the exception + message headers to raygun.

### DIFF
--- a/src/ServiceInsight/Framework/Commands/CommandsModule.cs
+++ b/src/ServiceInsight/Framework/Commands/CommandsModule.cs
@@ -6,13 +6,10 @@ namespace ServiceInsight.Framework.Commands
     {
         protected override void Load(ContainerBuilder builder)
         {
-            builder.RegisterType<CopyConversationIDCommand>().SingleInstance();
-            builder.RegisterType<CopyMessageURICommand>().SingleInstance();
-            builder.RegisterType<RetryMessageCommand>().SingleInstance();
-            builder.RegisterType<SearchByMessageIDCommand>().SingleInstance();
-            builder.RegisterType<ShowSagaCommand>().SingleInstance();
-            builder.RegisterType<ChangeSelectedMessageCommand>().SingleInstance();
-            builder.RegisterType<ShowExceptionCommand>().SingleInstance();
+            builder.RegisterAssemblyTypes(GetType().Assembly)
+                .Where(t => t.BaseType == typeof(BaseCommand))
+                .AsSelf()
+                .SingleInstance();
         }
     }
 }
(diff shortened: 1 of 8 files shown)


johnsimons:
@distantcam what is displayed after "... message data: " ?

johnsimons:
Is that an error message ?

johnsimons:
Are we showing the user exactly what data we are sending to us ?

distantcam:
@johnsimons Yes that's the exception message.

No we are not showing what data is being sent to us. Do we need to?

johnsimons:
@distantcam so won't that message always be the same ?

distantcam:
@johnsimons Nope

johnsimons:
@distantcam 

> No we are not showing what data is being sent to us. Do we need to?

What is your gut feeling ? Would u press that button without knowing exactly what is being sent ? Also do we have user contact details if we need more info ?

johnsimons:
@distantcam can u give me some examples of what the message is ?

distantcam:
@johnsimons 

> What is your gut feeling ? Would u press that button without knowing exactly what is being sent ? Also do we have user contact details if we need more info ?

It's the same as the exception report dialog, plus the headers of the message. I'd press the button, and if I was concerned I'd watch the packet with fiddler first.

> can u give me some examples of what the message is ?

So far we have 
- Only one arrow is allowed to come in
- The given key was not present in the dictionary.

johnsimons:
> It's the same as the exception report dialo

pretty sure @gbiellem raised an issue to fix this, so we do give the option for the user to see the data that we are about to send. @gbiellem do u remember this ?

johnsimons:
> The given key was not present in the dictionary.

That is not a very nice thing to display to the user.

How about all we say is:
There was a problem parsing the data to display the diagram.
Click here ....

distantcam:
> That is not a very nice thing to display to the user.

I agree, but some users have a bad habit of screenshotting the error message and sending that instead of the nice fancy report error button. At least this way we'll have something. :/

johnsimons:
@distantcam if a user reports a screenshot, that is IMO a good thing, because hopefully we can contact them and ask more questions.
So I do think that displaying a generic message to the user is in this scenario a better option

johnsimons:
BTW, that is what we have decided previously, see https://github.com/Particular/ServiceInsight/issues/545#issuecomment-168962919

distantcam:
I'm happy to change the message to a more generic one, but I would like to keep the report error feature as this will cover other errors the diagram might have, not just #545

johnsimons:
Yes of course, see https://github.com/Particular/ServiceInsight/issues/545#issuecomment-169159023

distantcam:
Well I had read #545 prior to creating this PR. ;)

HEskandari:
Would have been good to return a 'ticket number' to the client, so that they can follow up the issue with us using that...maybe we could generate a GUID (or some sort of a unique-hash of the report) and give it back to the client while including that in the report?

PR #559, opened 2016-02-03
Truncate saga update processing

distantcam:
For long running sagas, too many updates causes ServiceInsight to hang trying to process the messages.

This truncates the list of updates to the first 100 and last 100 and places a message in the Saga view saying the middle data is unavailable.

### DIFF
--- a/src/ServiceInsight/Framework/MessageDecoders/JsonMessageDeserializer.cs
+++ b/src/ServiceInsight/Framework/MessageDecoders/JsonMessageDeserializer.cs
@@ -6,6 +6,7 @@
     using System.Globalization;
     using System.Linq;
     using System.Text;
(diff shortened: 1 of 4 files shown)


sergioc:
> places a message in the Saga view saying the middle data is unavailable.

Is that the real reason or are we doing this because there is too much data? Can the message be changed to something like:

> **Saga data too large.**
Because of the large amount of data in this saga, ServiceInsight is only able to show the first 100 and the last 100 messages.

distantcam:
@sergioc I like that wording. I'll go change the message.

HEskandari:
In the context of "SagaUpdate" wouldn't this be better?

"Because of the large amount of data in this saga, ServiceInsight is only able to show the first 100 and the last 100 **changes**."

Issue #564, opened 2016-02-04
ArgumentOutOfRangeException: Specified argument was out of the range of valid values. 


distantcam:
Reported by a user from Raygun. Here's the stack trace.

Particular.ServiceInsight.Desktop.MessageList.MessageErrorInfo.ToString() in c:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\MessageList\MessageErrorInfo.cs:104
 DevExpress.Xpf.Editors.EditStrategyBase.FormatDisplayTextFast(Object editValue):10
 DevExpress.Xpf.Editors.EditStrategyBase.FormatDisplayTextInternal(Object editValue, Boolean applyFormatting):36
 DevExpress.Xpf.Editors.EditStrategy.TextEditStrategy.FormatDisplayText(Object editValue, Boolean applyFormatting):27
 DevExpress.Xpf.Editors.BaseEdit.FormatDisplayText(Object editValue, Boolean applyFormatting):13
 DevExpress.Xpf.Editors.BaseEdit.GetDisplayText(Object editValue, Boolean applyFormatting):0
 DevExpress.Xpf.Editors.BaseEdit.DevExpress.Xpf.Editors.IBaseEdit.GetDisplayText(Object editValue, Boolean applyFormatting):3
 DevExpress.Xpf.Editors.Settings.BaseEditSettings.GetDisplayText(Object editValue, Boolean applyFormatting):0
 DevExpress.Xpf.Editors.Settings.TextEditSettings.GetDisplayTextFromEditor(Object editValue):36
 DevExpress.Xpf.Grid.DataViewBase.GetDisplayObject(Object value, ColumnBase column, Boolean applyFormat):15
 DevExpress.Xpf.Grid.DataViewBase.GetColumnDisplayText(Object value, ColumnBase column, Nullable`1 rowHandle):9
 DevExpress.Xpf.Grid.DataViewBase.GetRowCellDisplayText(Int32 rowHandle, Int32 visibleColumnIndex):89
 DevExpress.Xpf.Grid.DataViewBase.GetTextForClipboard(Int32 rowHandle, Int32 visibleColumnIndex):8
 DevExpress.Xpf.Grid.Native.TableViewBehavior.GetDataRowText(StringBuilder sb, Int32 rowHandle):18
 DevExpress.Xpf.Grid.RowsClipboardController.GetTextInRows(IEnumerable`1 rows):103
[...]

PR #565, opened 2016-02-04
Better exception message for enum

distantcam:
Fix for #564 

This will change the error reports we get to have more info on what the enum value is.


distantcam:
Also added a change to include the ServiceControl version in the bug report.

Issue #566, opened 2016-03-03
Raygun NullReferenceException


distantcam:
NullReferenceException reported via Raygun for version 1.5.4+15.Branch.master.Sha.3a70dbbf7020d010778cb4b09542d09b33eadf18

```
ServiceInsight.Shell.ShellView.OnRestoreLayout(ISettingsProvider settingsProvider) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Shell\Shell.xaml.cs:85
 ServiceInsight.Shell.ShellViewModel.RestoreLayout() in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Shell\ShellViewModel.cs:147
 Caliburn.Micro.ViewAware.Caliburn.Micro.IViewAware.AttachView(Object view, Object context):58
 Caliburn.Micro.ViewModelBinder.<.cctor>b__3(Object viewModel, DependencyObject view, Object context):170
 Caliburn.Micro.WindowManager.CreateWindow(Object rootModel, Boolean isDialog, Object context, IDictionary`2 settings):22
 Caliburn.Micro.WindowManager.ShowWindow(Object rootModel, Object context, IDictionary`2 settings):53
 ServiceInsight.Startup.AppBootstrapper.OnStartup(Object sender, StartupEventArgs e) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Startup\AppBootstrapper.cs:111
 System.Windows.Application.OnStartup(StartupEventArgs e):31
 ServiceInsight.App.OnStartup(StartupEventArgs e) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\App.xaml.cs:54
 System.Windows.Application.<.ctor>b__1_0(Object unused):15
 System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs):119
 System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler):2
```

distantcam:
Here's a stacktrace from 1.4.1+2.Branch.master.Sha.eab6d38808e52e5f757ae73f1480b280121ab0c1

```
[...]

PR #567, opened 2016-03-08
RestoreLayout NullRef fix

distantcam:
I can't replicate the cause of this, but once the settings have saved `null` as the value, the restore layout breaks exactly as the issue describes.

This fixes the issue when a null layout is saved.

See #566

### DIFF
--- a/src/ServiceInsight/Framework/Settings/SettingsProvider.cs
+++ b/src/ServiceInsight/Framework/Settings/SettingsProvider.cs
@@ -21,7 +21,9 @@ namespace ServiceInsight.Framework.Settings
             {
                 if (settingsRepository.HasSettings(GetKey<T>()))
                 {
-                    return LoadSettings<T>(ReadSettingMetadata<T>());
+                    var settings = LoadSettings<T>(ReadSettingMetadata<T>());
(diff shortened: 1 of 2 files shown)


distantcam:
What if there was a situation where it's valid to store `null` as a value for a setting? By changing the `SettingsProvider` we can no longer store null for any setting, not just the layout which is the current issue.

HEskandari:
The ```SettingsProvider``` manages that objects that contain the settings value(s). The values can be NULL if need be, the setting holder instance can not, e.g.

```settingsProvider.GetSettings<ShellLayoutSettings>()``` should never be null. That's exactly what you are doing in the PR:

``` var layoutSetting = settingProvider.GetSettings<ShellLayoutSettings>() ?? new ShellLayoutSettings();```

All I'm suggesting is that instead of doing it only for "ShellLayoutSettings" do it across the board, and move it in the Settings Provider class. It is already doing this if there are exceptions thrown.

distantcam:
Ok. I was just raising the idea of the holder being null, but I'm happy to say that shouldn't be allowed. In that case yeah the settings provider should deal with this. I'll make the changes.

Issue #571, opened 2016-03-21
Unable to cast object of type 'ServiceInsight.ServiceControl.PresentationBody' to type 'System.Tuple`2[System.String,ServiceInsight.ServiceControl.PresentationBody]'.


ramonsmits:
By clicking rapidly between messages it seems that an InvalidCastException can be raised.

Exception:
>System.InvalidCastException: Unable to cast object of type  ServiceInsight.ServiceControl.PresentationBody' to type 'System.Tuple`2[System.String,ServiceInsight.ServiceControl.PresentationBody]'

Exception details:

```
System.InvalidCastException: Unable to cast object of type  ServiceInsight.ServiceControl.PresentationBody' to type 'System.Tuple`2[System.String,ServiceInsight.ServiceControl.PresentationBody]'.
at ServiceInsight.ServiceControl.DefaultServiceControl.Execute[T](RestRequestWithCache request, Func`2 selector, String baseUrl) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\ServiceControl\DefaultServiceControl.cs:line 280
at ServiceInsight.ServiceControl.DefaultServiceControl.LoadBody(StoredMessage message) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\ServiceControl\DefaultServiceControl.cs:line 168
at ServiceInsight.MessageViewers.MessageBodyViewModel.LoadMessageBody() in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\MessageViewers\MessageBodyViewModel.cs:line 160
```

Location in source code:

-  https://github.com/Particular/ServiceInsight/blob/master/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs#L280

It seems that

- When `CacheStyle.Immutable` then the cache should contain that item directly and it its `CacheStyle.IfNotModified` than the type should be wrapped in a `Tuple<string,T>`.
- The `CacheStyle` can be different for different instance of the same type (`PresentationBody`).
- In this issue it is assumed that `PresentationBody` has cache style `IfNotModified` but seems to have been added as `Immutable`.

Versions:

- ServiceInsight: 1.5.4
- ServiceControl: 1.11.1

References

- https://nservicebus.desk.com/agent/case/17706

HEskandari:
@ramonsmits regarding your observations after going through the code in more depth:

[...]

PR #573, opened 2016-03-22
Always cache message body

HEskandari:
caching regardless of the message state, as it won't change on retries.


HEskandari:
This is related to issue #571

ramonsmits:
@HEskandari You changed it to immutable, what if a message is retried? Will that be a 'new' message?Will the user see the updated headers?

HEskandari:
@ramonsmits this is only the Body of the message (payload), it will not affect loading of the header or other aspects.

PR #581, opened 2016-05-09
Add install info for SagaAudit v6

boblangley:
Connects to https://github.com/Particular/PlatformDevelopment/issues/677

### DIFF
--- a/src/ServiceInsight/Saga/SagaWindowView.xaml
+++ b/src/ServiceInsight/Saga/SagaWindowView.xaml
@@ -20,7 +20,7 @@
                     <DataTemplate DataType="{x:Type l:SagaUpdate}">
                         <l:SagaUpdateControl DataContext="{Binding}"
                                              SelectedMessageId="{Binding DataContext.SelectedMessageId,
-                                                                       ElementName=uc}"
+                                                                         ElementName=uc}"
                                              TimeoutClick="Hyperlink_Click" />
                     </DataTemplate>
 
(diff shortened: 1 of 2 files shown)


HEskandari:
@boblangley do we need to display the message for both v5 and v6? Could we simplify the message and copy the right command (on CopyCommand) based on the version?

boblangley:
@HEskandari  We display the message for both v4 and v5 so I just assumed we would do the same for v6. If we want to handle this differently I am not opposed.

HEskandari:
This is what I'm proposing:

image

What do you think @distantcam?

Do you mind if I update the PR (if/once/when this is approved)?

boblangley:
Not at all.

HEskandari:
@boblangley what's your thoughts on the proposed change?

boblangley:
I am all for it and appreciate the help.

distantcam:
I don't like having a generic `x` there. Could we have the text update the appropriate version based on the selected message?

HEskandari:
@distantcam Don't think there'd be any issues with that. That'd actually be even better.

HEskandari:
@boblangley @distantcam Please review.

boblangley:
@HEskandari Looks good to me. I ran the v6 SagaAudit plugin smoke test sending the control messages to the wrong queue. The Saga tab works as expected. :+1:

boblangley:
Should we have a fallback if the version header is missing or unparseable? Are we confident about the version information being in the headers?

HEskandari:
@boblangley AFAIK version has been there for quite some time. Think @johnsimons might be able to confirm?

johnsimons:
Yes version has been there since the beggining

HEskandari:
@boblangley Merged. Do we need to get a release out as a part of v6 launch or prior?

boblangley:
I think as part of is good enough.

HEskandari:
Is there a separate task/item to track the release? Should we close this issue?

boblangley:
Yes let's close the pr. We can track the release to support v6 in another issue for a later stage. Unless of course we end up doing a Service Insight release in the meantime.

Issue #580, opened 2016-05-05
Look for a valid license on multiple locations (HKCU and HKLM)


HEskandari:
The new licensing package would look for a license in both HKCU and HKLM. If you have an expired or trial license at one location, the licensing module will still look for a valid license in the other location.

HEskandari:
Tested this and it worked fine. Once we merge [this PR](https://github.com/Particular/ServiceInsight/pull/589), we should be able to release it.

PR #588, opened 2016-06-21
Licensing Update

gbiellem:
Relates to #580 

This resolves the 32 and 64 bit reg issue and removes the issue where an invalid license in HKCU would trump a valid license in HKLM


gbiellem:
@HEskandari  here's the licensing update we discussed in slack.  Please review

HEskandari:
@gbiellem the code looks good as I have reviewed it before too.

PR #599, opened 2016-08-18
Ordering of Handlers based on HandledAt property to ensure the right ordering

### LINKED ISSUE #570: Sequence Diagram misplaces Handlers
## Symptoms

The Sequence Diagram misplaces the handler boxes in occasions and that results the message to appear out of order.

screen1
## Who's affected

This update is recommended for all users of ServiceInsight who rely on Sequence Diagram as it draws a more accurate sequence of messages in certain situations.

HEskandari:
Resolves #570


distantcam:
Looks good :+1:

PR #600, opened 2016-09-03
Introduced rx sc component and adapted endpoint explorer

distantcam:
I've created a reactive ServiceControl interface, which changes the interaction with SC from a pull to a push model. The main benefit here is a single call to SC which then broadcasts the data within SI rather than each viewmodel calling SC independently, which is causing SI to call SC multiple times.

The second part of this change is to rewrite the Endpoint Explorer to use this new interface. This lead to some code simplification.

### DIFF
--- a/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
+++ b/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
@@ -22,14 +22,14 @@
         IEventAggregator eventAggregator;
         string initialEndpoint;
 
-        public EndpointExplorerViewModel(IEventAggregator eventAggregator, CommandLineArgParser commandLineParser)
+        public EndpointExplorerViewModel(IRxServiceControl serviceControl, IEventAggregator eventAggregator, CommandLineArgParser commandLineParser)
         {
             this.eventAggregator = eventAggregator;
             Items = new BindableCollection<ExplorerItem>();
 
             initialEndpoint = commandLineParser.ParsedOptions.EndpointName;
 
-            RxServiceControl.Instance.Endpoints().Subscribe(MergeEndpoints);
+            serviceControl.Endpoints().Subscribe(MergeEndpoints);
         }
 
         private void MergeEndpoints(ServiceControlData e)
--- a/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
+++ b/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Linq;
     using Caliburn.Micro;
-    using EndpointExplorer;
     using Framework;
     using Framework.Events;
     using Newtonsoft.Json;
@@ -11,11 +10,11 @@
 
     public class EndpointExplorerViewModel : Screen, IHandle<RequestSelectingEndpoint>, IHandle<SelectedExplorerItemChanged>
     {
-        static JsonSerializer serializer = new JsonSerializer();
+        static JsonSerializer serializer;
 
         static EndpointExplorerViewModel()
         {
-            serializer.ContractResolver = new SnakeCasePropertyNamesContractResolver();
+            serializer = new JsonSerializer { ContractResolver = new SnakeCasePropertyNamesContractResolver() };
         }
 
         IEventAggregator eventAggregator;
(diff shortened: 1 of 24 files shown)


src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
```diff
@@ -3,203 +3,97 @@
...
+        static EndpointExplorerViewModel()
         {
-            this.eventAggregator = eventAggregator;
-            this.settingsProvider = settingsProvider;
-            this.serviceControl = serviceControl;
-            this.networkOperations = networkOperations;
-            this.connectionProvider = connectionProvider;
-            this.commandLineParser = commandLineParser;
-            Items = new BindableCollection<ExplorerItem>();
+            serializer.ContractResolver = new SnakeCasePropertyNamesContractResolver();
```
HEskandari: I find the split a bit confusing...could this be merged with the initialiation code above? e.g.

    static EndpointExplorerViewModel()
    {
        serializer = new JsonSerializer { ContractResolver = new SnakeCasePropertyNamesContractResolver() }; 
    }

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
```diff
@@ -3,203 +3,97 @@
...
-        }
+                Items.Add(root);
+            }
 
-        string GetConfiguredAddress()
-        {
-            if (commandLineParser.ParsedOptions.EndpointUri == null)
+            var toRemove = root.Children.ToList();
+            var endpointInstancesGroupedByName = e.Data.OrderBy(x => x.name).GroupBy(x => x.name);
+            foreach (var scaledOutEndpoint in endpointInstancesGroupedByName)
```
HEskandari: hmmm why is this called scaledOutEndpoint?

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
```diff
@@ -3,203 +3,97 @@
...
-        void SelectDefaultEndpoint()
+        public void Handle(RequestSelectingEndpoint message)
         {
-            if (ServiceControlRoot == null)
+            foreach (var item in Items.OfType<ServiceControlExplorerItem>())
             {
-                return;
-            }
-
-            if (!commandLineParser.ParsedOptions.EndpointName.IsEmpty())
```
HEskandari: This code was required for when SI is activated via a link, e.g. si://ServiceControlUrl/endpoint and it'd focus on that particular endpoint. Was it removed intentionally?

src/ServiceInsight/Framework/RxServiceControl.cs
```diff
@@ -0,0 +1,181 @@
...
+        void SetRefresh(TimeSpan interval);
+
+        void DisableRefresh();
+    }
+
+    class RxServiceControl : IRxServiceControl
+    {
+        static Serilog.ILogger anotarLogger = Log.ForContext<IRxServiceControl>();
+
+        public static IRxServiceControl Instance { get; set; }
```
HEskandari: hmmmm isn't this registered as Singleton in the container? why do we need another singleton on top of that?

src/ServiceInsight/Startup/AppBootstrapper.cs
```diff
@@ -27,6 +28,10 @@ public AppBootstrapper()
 
         protected override void Configure()
         {
+            BlobCache.ApplicationName = "ServiceInsight";
+
+            RxServiceControl.Instance = new RxServiceControl(BlobCache.UserAccount);
```
HEskandari: Same point as above...why reimplement singleton?

HEskandari:
Made some comments. The only bigger reservation I have is how `JObject`s are passed around and scattered and the new `IRxServiceControl` although has fine grained methods (GetMessages, GetEndpoints) the results are not typed DTOs and are `dynamic` or `JObject`s.

src/ServiceInsight/Framework/RxServiceControl.cs
distantcam: No it's not registered as a singleton in the container.

I feel that with singletons we should start to take them out of the container to simplify the constructors for certain objects. I'm not sure about this. We should probably have a good discussion about it.

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
distantcam: No this was unintentional. I forgot to check that this functionality remained. Let me check and see if I need to update this PR.

distantcam:
> The only bigger reservation I have is how JObjects are passed around and scattered and the new IRxServiceControl although has fine grained methods (GetMessages, GetEndpoints) the results are not typed DTOs and are dynamic or JObjects.

Yep again, we should have a conversation about this.

My point of view is this, yes I used to be in the same boat as you, which is to convert things to a POCO. Over time however, those POCO's change. Common values are put into a base class, and helper methods are added, and suddenly we no longer have a POCO. Also why convert to a POCO? It's a layer between the data and the code that works on the data. What does the POCO give us for that layer?

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
distantcam: Fixed this issue now

PR #606, opened 2016-10-24
Removing Caliburn.Micro and using Pirac

distantcam:


adamralph:
We are reverting master to 1.5.7 in #617 so these changes will no longer be in master (but they will be retained in the history). We are going to reevaluate our options regarding changing platforms/frameworks before revisiting this.

PR #612, opened 2016-10-31
Fixes #596.

### LINKED ISSUE #596: JSON Message Body is not displayed correctly, if the message contains HTML
The JSON formatting gets messed up, when the message contains HTML tags:

untitled

HEskandari:
Removes the unnecessary parsing of quotes and backslashes from the SC response. Refactors how the BOM header is cleaned up.

---

Fixes #596

### DIFF
--- a/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
+++ b/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
@@ -6,6 +6,7 @@
     using System.Linq;
     using System.Net;
     using System.Runtime.Caching;
+    using System.Text;
     using System.Xml;
     using System.Xml.Linq;
     using Anotar.Serilog;
@@ -140,7 +141,7 @@
 
             return body.StartsWith("<?xml") ?
                 GetXmlData(body) :
-                JsonPropertiesHelper.ProcessValues(body, CleanupBodyString);
+                JsonPropertiesHelper.ProcessValues(body, ResponseUnicodeCleaner.CleanUp);
         }
 
         public void LoadBody(StoredMessage message)
@@ -160,7 +161,7 @@
                 switch (response.StatusCode)
                 {
                     case HttpStatusCode.OK:
-                        presentationBody.Text = CleanupBodyString(response.Content);
+                        presentationBody.Text = ResponseUnicodeCleaner.CleanUp(response.Content);
                         break;
                     case HttpStatusCode.NoContent:
                         presentationBody.Hint = PresentationHint.NoContent;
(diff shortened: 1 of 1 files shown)


HEskandari:

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -533,4 +532,14 @@ public PresentationHint Hint
             set { hint = value; }
         }
     }
+
+    public class ResponseUnicodeCleaner
```
distantcam: Do we need a whole new class just for this code?

HEskandari: What's the problem with a 'whole new class'? Cleaning unicode characters is not a responsibility of the ServiceControl class. Moving it out in favor of SRP and cleaner code. Happy to extract the class and move it to a different namespace though, if it helps.

distantcam: > Cleaning unicode characters is not a responsibility of the ServiceControl class

It can be if communicating with ServiceControl requires it. The ServiceControl class is responsible for communicating, and should hide the details.

I don't like having a class with so little code in it too, but I'll admit that's a personal preference. Also currently the code isn't being reused so I prefer the method being inside the class using it.

adamralph: Without knowing too much about the `DefaultServiceControl` class, my instinct would be to have a private method on `DefaultServiceControl` to do this work rather than a new class.

From my naive POV `DefaultServiceControl` does seem to be doing a lot but if I were to break it apart I'd probably start top down rather than bottom up.

HEskandari:
@Particular/serviceinsight-maintainers updated.

distantcam:
> Is it possible to write a test for this?

@adamralph It is, but it's one of those closed loop tests that wouldn't catch this bug anyway. To test we'd need to feed fake data into the class, and then test that the class processed the fake data correctly. That won't help when the data is not what we anticipated.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -510,6 +510,11 @@ void LogError(IRestResponse response)
             LogTo.Error(exception, errorMessage);
         }
 
+        string CleanUp(string content)
+        {
+            return content.StartsWith(byteOrderMarkUtf8) ? content.Remove(0, byteOrderMarkUtf8.Length) : content;
+        }
+
```
adamralph: Only a minor cosmetic point, but it would have been nicer if the `CleanupBodyString` method above would have been replaced inline, conserving the method name, to isolate the changes in the diff.

adamralph:
@distantcam is this class already tested through that code path? If so, is the problem that the test does not contain the UTF8 pre-amble but production does? Or does production only contain it some cases?

PR #614, opened 2016-10-31
Fixes #613.

### LINKED ISSUE #613: NRE when auto-connecting to a ServiceControl instance that is no longer available
Steps to reproduce:

1. Start SI and connect to an instance of SC
2. Exit SI
3. Stop or remove that instance 
4. Start SI

Upon startup SI tries to connect to the previously connected instance, but it throws an exception. The application crashes on Splash screen.

HEskandari:
Doing a null check on the url SI connects to. Also doing that on OnInitialize to fix the auto-connect on startup issue.

### DIFF
--- a/src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
+++ b/src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
@@ -15,6 +15,11 @@
         {
             Url = url;
 
+            if (url == null)
+            {
+                return;
+            }
+
             AsyncPump.Run(async () =>
             {
                 await serviceControl.ClearServiceControls();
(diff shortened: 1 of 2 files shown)


HEskandari:

adamralph:
@HEskandari why was this tagged as being blocked?

src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
```diff
@@ -15,6 +15,11 @@ public void ConnectTo(string url)
         {
             Url = url;
 
+            if (url == null)
```
adamralph: Would it be better to throw an exception in this case? I guess we're fixing the problem at the call site below?

HEskandari: Don't think so. This is called on application start as well which means if an exception is thrown the app will crash.

PR #615, opened 2016-11-01
[WIP] Add license file

johnsimons:


LICENSE
```diff
@@ -0,0 +1,10 @@
+By accessing NServiceBus code here, you are agreeing to the following licensing terms.
```
HEskandari: Should this be renamed to ServiceInsight?

LICENSE
```diff
@@ -0,0 +1,10 @@
...
+If you do not agree to these terms, do not access the NServiceBus code.
+
+Your license to the NServiceBus source and/or binaries is governed by the Reciprocal Public License 1.5 (RPL1.5) license as described here:
```
HEskandari: Same as above.

LICENSE
johnsimons: Not really, we are registered as NServiceBus

distantcam: Registered?

johnsimons: Yes officially we are NServiceBus doing business as Particular Software

johnsimons:
Trying to clarify, see https://github.com/Particular/PlatformDevelopment/issues/983#issuecomment-257498951

LICENSE
HEskandari: This makes sense when read off NServiceBus repo, but not ServiceInsight and / or ServicePulse. If 'NServiceBus' referred to here means the company, could it be clarified? I mean, this is a license that people would need to read and understand clearly so why leave it vague?

johnsimons: https://github.com/Particular/PlatformDevelopment/issues/983#issuecomment-257498951

adamralph:
Added to 1.5.9 milestone.

PR #616, opened 2016-11-02
[WIP] Made SC connection provider a singleton

### LINKED ISSUE #574: Connecting to Different Service Control Instance Does Nothing
v1.5.6

Connecting to a different ServiceControl instance DOES NOTHING (until ServiceInsight is restarted or refreshed). This results in a bad user experience when switching between Dev, Test, and Prod, environments:
1. User connects to a different SC instance and clicks OK.
2. User waits a long time for the change to take effect.
3. User eventually gives up waiting and restarts SI or clicks Refresh.

Please fix this. You could alternatively update the "Connect to ServiceControl" dialog to say something like: "You must click Refresh for this change to take effect", but that sure feels like a hack. 

Thank you,
Danny

distantcam:
This fixes #574

- ServiceControlConnectionProvider was not a singleton, so each VM had a different instance and when one was updated with the new connection, the others were not updated.
- Changed the EndpointExplorer to replace the root node, which was the partial implementation from before.

### DIFF
--- a/src/ServiceInsight/Framework/Modules/CoreModule.cs
+++ b/src/ServiceInsight/Framework/Modules/CoreModule.cs
@@ -21,14 +21,16 @@
             builder.RegisterType<HeaderContentDecoder>().As<IContentDecoder<IList<HeaderInfo>>>();
             builder.RegisterType<NetworkOperations>().SingleInstance();
             builder.RegisterType<AppLicenseManager>().SingleInstance();
-            builder.RegisterType<ServiceControlConnectionProvider>().SingleInstance();
-            builder.RegisterType<DefaultServiceControl>().As<IServiceControl>().InstancePerLifetimeScope();
             builder.RegisterInstance(new RxServiceControl(BlobCache.UserAccount)).As<IRxServiceControl>().SingleInstance();
             builder.RegisterType<CommandLineArgParser>().SingleInstance().OnActivating(e => e.Instance.Parse());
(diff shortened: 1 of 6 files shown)


src/ServiceInsight/Framework/Modules/CoreModule.cs
```diff
@@ -21,7 +21,7 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<HeaderContentDecoder>().As<IContentDecoder<IList<HeaderInfo>>>();
             builder.RegisterType<NetworkOperations>().SingleInstance();
             builder.RegisterType<AppLicenseManager>().SingleInstance();
-            builder.RegisterType<ServiceControlConnectionProvider>().InstancePerLifetimeScope();
+            builder.RegisterType<ServiceControlConnectionProvider>().SingleInstance();
```
HEskandari: Just wondering...isn't InstancePerLifetimeScope essentially the same as singleton here? Since we only have one container that is never disposed, effectively InstancePerLifetimeScope would give us singletons...anything I'm missing?

distantcam: Turns out we are using lifetime scoping, so InstancePerLifetimeScope is correct, and the real issue was elsewhere.

adamralph:
Unassigning until this is reconsidered after the merge of #617

distantcam:
Closing due to #617 and reimplementing

PR #621, opened 2016-11-09
Refactored WorkStarted/WorkFinished events into a WorkNotifier service.

distantcam:
This change bundles WorkStarted and WorkFinished notifications into a disposable, to ensure that the WorkFinished event is always sent out. This starts to sort out problems where a WorkFinished event would occur before a WorkStarted, or even without a matching WorkStarted.

### DIFF
--- a/src/ServiceInsight.Tests/MessageListViewModelTests.cs
+++ b/src/ServiceInsight.Tests/MessageListViewModelTests.cs
@@ -19,6 +19,7 @@
     public class MessageListViewModelTests
     {
         IEventAggregator eventAggregator;
+        IWorkNotifier workNotifier;
         IServiceControl serviceControl;
         SearchBarViewModel searchBar;
         Func<MessageListViewModel> messageListFunc;
@@ -28,15 +29,18 @@
         public void TestInitialize()
         {
             eventAggregator = Substitute.For<IEventAggregator>();
+            workNotifier = Substitute.For<IWorkNotifier>();
             serviceControl = Substitute.For<IServiceControl>();
             searchBar = Substitute.For<SearchBarViewModel>();
             clipboard = Substitute.For<IClipboard>();
-            messageListFunc = () => new MessageListViewModel(eventAggregator,
-                                                   serviceControl,
-                                                   searchBar,
-                                                   Substitute.For<GeneralHeaderViewModel>(),
-                                                   Substitute.For<MessageSelectionContext>(),
-                                                   clipboard);
+            messageListFunc = () => new MessageListViewModel(
+                eventAggregator,
+                workNotifier,
+                serviceControl,
+                searchBar,
+                Substitute.For<GeneralHeaderViewModel>(),
+                Substitute.For<MessageSelectionContext>(),
+                clipboard);
         }
 
         [Test]
@@ -44,23 +48,21 @@
         {
             var endpoint = new Endpoint { Host = "localhost", Name = "Service" };
             serviceControl.GetAuditMessages(Arg.Is(endpoint), Arg.Any<string>(), Arg.Any<int>(), Arg.Any<string>(), Arg.Any<bool>())
-                             .Returns(x => new PagedResult<StoredMessage>
-                             {
-                                 CurrentPage = 1,
-                                 TotalCount = 100,
-                                 Result = new List<StoredMessage>
-                                 {
-                                     new StoredMessage(),
-                                     new StoredMessage()
-                                 }
-                             });
+                .Returns(x => new PagedResult<StoredMessage>
+                {
+                    CurrentPage = 1,
+                    TotalCount = 100,
+                    Result = new List<StoredMessage>
+                    {
+                        new StoredMessage(),
+                        new StoredMessage()
+                    }
+                });
 
             var messageList = messageListFunc();
 
             messageList.Handle(new SelectedExplorerItemChanged(new AuditEndpointExplorerItem(endpoint)));
 
-            eventAggregator.Received(1).PublishOnUIThread(Arg.Any<WorkStarted>());
-            eventAggregator.Received(1).PublishOnUIThread(Arg.Any<WorkFinished>());
             messageList.Rows.Count.ShouldBe(2);
             searchBar.IsVisible.ShouldBe(true);
         }
--- a/src/ServiceInsight/Framework/Modules/CoreModule.cs
+++ b/src/ServiceInsight/Framework/Modules/CoreModule.cs
@@ -21,6 +21,7 @@
             builder.RegisterType<ServiceControlConnectionProvider>().InstancePerLifetimeScope();
             builder.RegisterType<DefaultServiceControl>().As<IServiceControl>().InstancePerLifetimeScope();
             builder.RegisterType<CommandLineArgParser>().SingleInstance().OnActivating(e => e.Instance.Parse());
+            builder.RegisterType<WorkNotifier>().As<IWorkNotifier>();
         }
     }
 }
--- a/src/ServiceInsight/Framework/Modules/CoreModule.cs
+++ b/src/ServiceInsight/Framework/Modules/CoreModule.cs
@@ -21,7 +21,7 @@
             builder.RegisterType<ServiceControlConnectionProvider>().InstancePerLifetimeScope();
             builder.RegisterType<DefaultServiceControl>().As<IServiceControl>().InstancePerLifetimeScope();
             builder.RegisterType<CommandLineArgParser>().SingleInstance().OnActivating(e => e.Instance.Parse());
-            builder.RegisterType<WorkNotifier>().As<IWorkNotifier>().SingleInstance();
+            builder.RegisterType<WorkNotifier>().As<IWorkNotifier>().InstancePerLifetimeScope();
         }
     }
 }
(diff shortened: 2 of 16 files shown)


HEskandari:
@distantcam can we have a bug created for this as well?

src/ServiceInsight.Tests/MessageListViewModelTests.cs
```diff
@@ -28,39 +29,40 @@ public class MessageListViewModelTests
...
+                        new StoredMessage()
+                    }
+                });
 
             var messageList = messageListFunc();
 
             messageList.Handle(new SelectedExplorerItemChanged(new AuditEndpointExplorerItem(endpoint)));
 
-            eventAggregator.Received(1).PublishOnUIThread(Arg.Any<WorkStarted>());
-            eventAggregator.Received(1).PublishOnUIThread(Arg.Any<WorkFinished>());
```
adamralph: Should we preserve these assertions?

I [gave it a try](https://github.com/adamralph/ServiceInsight/commit/9356355b13826d8aeb32c4444504bd8e0fbedca7) and the test passes. The assertions on `workNotifier` and `notifyOfWork` are the equivalent of these two lines.

I'm not sure how important it is to test this interaction. I usually favour testing at the black box level first and drilling deeper only when required, but I don't know if this is already covered by any higher level test or whether it's practical to do so.

src/ServiceInsight.Tests/MessageListViewModelTests.cs
```diff
@@ -28,39 +29,40 @@ public class MessageListViewModelTests
...
+                .Returns(x => new PagedResult<StoredMessage>
+                {
+                    CurrentPage = 1,
+                    TotalCount = 100,
+                    Result = new List<StoredMessage>
+                    {
+                        new StoredMessage(),
+                        new StoredMessage()
+                    }
+                });
```
HEskandari: Could we change the return result so it makes sense? There `100` total count is clearly incorrect!

src/ServiceInsight/Framework/Modules/CoreModule.cs
```diff
@@ -21,6 +21,7 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<ServiceControlConnectionProvider>().InstancePerLifetimeScope();
             builder.RegisterType<DefaultServiceControl>().As<IServiceControl>().InstancePerLifetimeScope();
             builder.RegisterType<CommandLineArgParser>().SingleInstance().OnActivating(e => e.Instance.Parse());
+            builder.RegisterType<WorkNotifier>().As<IWorkNotifier>();
```
HEskandari: This can be registered as Singleton, since the WorkNotifier itself is not disposable (it generates disposable objects).

src/ServiceInsight/Framework/WorkNotifier.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        IDisposable NotifyOfWork(string startMessage);
+
+        IDisposable NotifyOfWork(string startMessage, string finishMessage);
+    }
+
+    class WorkNotifier : IWorkNotifier
+    {
+        IEventAggregator eventAggregator;
+
```
HEskandari: Does it make sense to merge `IWorkTracker` functionality into this and have one place to deal with work stop / start / status?

src/ServiceInsight.Tests/MessageListViewModelTests.cs
distantcam: @adamralph Good point. I'll try to do that from now on.

@HEskandari I'm not sure I want to change that. The fact that the count is different from the actual result means if we ever assume they are the same this test will fail, which means our assumption is wrong. I like this.

src/ServiceInsight.Tests/MessageListViewModelTests.cs
distantcam: @adamralph That's exactly why I removed this, it's assuming that the internal implementation uses WorkNotifier, which in turn uses EventAggregator. This isn't black box testing and would make the test brittle to changes in WorkNotifer and EventAggregator. Given the test is for making sure the view model calls the servicecontrol instance properly I didn't think also testing it calls WorkNotifier/EventAggregator as well was a good idea.

src/ServiceInsight/Framework/Modules/CoreModule.cs
distantcam: @HEskandari Is singleton the preferred option? I hadn't thought about that.

- Singleton means less objects for the GC to worry about
- Instance means not having to worry about multiple objects trying to change static state (if the object has any state).

src/ServiceInsight/Framework/WorkNotifier.cs
distantcam: Not all WorkTrackers are WorkNotifiers, and vice versa. One is a publisher, the other is a subscriber.

HEskandari: The WorkTracker interface tells you one thing that WorkNotifier is now managing: "Is there any work in progress"? To me it feels if WorkNotifier encapsulated its state, it'd know if there are work in progress or not and could answer that question. 

The subscriber would be the Shell where it just updates the UI with the messages where as now it is also answering if there are currently work in progress or not.

src/ServiceInsight/Framework/Modules/CoreModule.cs
adamralph: `WorkNotifier` is passed an `IEventAggregator` in its constructor and I see [`IEventAggregator` being registered with `InstancePerLifetimeScope`](https://github.com/Particular/ServiceInsight/blob/c52d649923798f2472a529fb37b57bd9f1c552a2/src/ServiceInsight/Framework/Modules/CaliburnModule.cs#L16).

If I understand correctly then the registration of `WorkNotifier` cannot be `SingleInstance`, but I guess it probably makes sense to increase its lifetime to `InstancePerLifetimeScope`.

src/ServiceInsight.Tests/MessageListViewModelTests.cs
adamralph: I'm fine to leave as is.

src/ServiceInsight/Framework/WorkNotifier.cs
distantcam: @HEskandari Firstly, the shell isn't the only thing that implements `IWorkTracker`.

I see your point, why even use the EventAggregator as a middle-man? Why not have ViewModels that care about work status "subscribe" directly to the WorkNotifier? We could, but that would be a bigger change. I'll keep that idea in mind though.

HEskandari: @distantcam agree that it is a bigger change. I'm happy with this as is, as it is a move in the right direction.

distantcam:
@adamralph I've made WorkNotifier a singleton. It works fine with IEventAggregator being InstancePerLifetimeScope.

Also outside the scope of this PR, but we should review the scope of everything registered into the container.

src/ServiceInsight/Framework/Modules/CoreModule.cs
```diff
@@ -21,6 +21,7 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<ServiceControlConnectionProvider>().InstancePerLifetimeScope();
             builder.RegisterType<DefaultServiceControl>().As<IServiceControl>().InstancePerLifetimeScope();
             builder.RegisterType<CommandLineArgParser>().SingleInstance().OnActivating(e => e.Instance.Parse());
+            builder.RegisterType<WorkNotifier>().As<IWorkNotifier>().SingleInstance();
```
adamralph: `WorkNotifier` takes an `IEventAggregator` in its constructor. The implementation of `IEventAggregator` is registered as `InstancePerLifetimeScope`. Does this not imply that `WorkNotifier` cannot be registered as `SingleInstance`?

Even if the *actual* behaviour of `IEventAggregator` is such that the container only ever creates one implementation, this is something that the code doesn't convey. I think it's better to register `WorkNotifier` as `InstancePerLifetimeScope` in this PR.

Then, if we want to change the implementations of both `IEventAggregator` and `IWorkNotifier` to `SingleInstance`, we can do that in a separate refactoring PR.

HEskandari: Since we don't create child containers or lifetime scopes, wouldn't the effective lifetime scope of `InstancePerLifetimeScope` would be Singleton? Unless we use `BeginLifetimeScope` to resolve them, they won't get disposed, according to [this](http://docs.autofac.org/en/latest/lifetime/index.html).

adamralph: > It's not about consistency it's about correct operation

Agreed, correct operation is the aim. Consistency in these two registrations is coincidental. Since `IEventAggregator` is registered as `InstancePerLifetimeScope`, `WorkNotifier` cannot be registered as `SingleInstance`. It has to be either `InstancePerLifetimeScope` or `InstancePerDependency`.

I also agree that we need to review all the container registrations as a separate exercise.

distantcam:
@adamralph I don't like rewriting history that much. So long as it's on a branch I don't care how many commits it has.

PR #623, opened 2016-11-10
Fixes #595

### LINKED ISSUE #595: Help does not point to ServiceInsight page
ServiceInsight Help goes to http://docs.particular.net but should be http://docs.particular.net/serviceinsight.

The PI and SC both deep link to the relevant doco, this should too.

HEskandari:
Just changed the link. 

fixes #595


adamralph:
Thanks @HEskandari !

PR #626, opened 2016-11-10
Resolves #574

### LINKED ISSUE #574: Connecting to Different Service Control Instance Does Nothing

HEskandari:
- Raising an event when connection to SC changes
- Clearing the message body views
- Clearing the saga view
- Clearing messages window.

fixes #574

### DIFF
--- a/src/ServiceInsight/MessageList/MessageListViewModel.cs
+++ b/src/ServiceInsight/MessageList/MessageListViewModel.cs
@@ -27,7 +27,8 @@
         IHandle<WorkFinished>,
         IHandle<AsyncOperationFailed>,
         IHandle<RetryMessage>,
-        IHandle<SelectedMessageChanged>
+        IHandle<SelectedMessageChanged>,
+        IHandle<ServiceControlConnectionChanged>
     {
         readonly IClipboard clipboard;
         IEventAggregator eventAggregator;
@@ -156,18 +157,23 @@
 
                 TryRebindMessageList(pagedResult);
 
-                SearchBar.IsVisible = true;
-                SearchBar.SetupPaging(new PagedResult<StoredMessage>
-                {
-                    CurrentPage = pagedResult.CurrentPage,
-                    TotalCount = pagedResult.TotalCount,
-                    Result = pagedResult.Result,
-                });
+                BindSearchBar(pagedResult);
             }
             finally
             {
                 eventAggregator.PublishOnUIThread(new WorkFinished());
             }
+        }
+
+        private void BindSearchBar(PagedResult<StoredMessage> pagedResult)
+        {
+            SearchBar.IsVisible = true;
+            SearchBar.SetupPaging(new PagedResult<StoredMessage>
+            {
+                CurrentPage = pagedResult.CurrentPage,
+                TotalCount = pagedResult.TotalCount,
+                Result = pagedResult.Result,
+            });
         }
 
         public MessageErrorInfo GetMessageErrorInfo(StoredMessage msg) => new MessageErrorInfo(msg.Status);
@@ -226,6 +232,11 @@
             }
         }
 
+        public void Handle(ServiceControlConnectionChanged message)
+        {
+            ClearResult();
+        }
+
         public void OnSelectedExplorerItemChanged()
         {
             RefreshMessages();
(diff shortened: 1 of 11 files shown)


HEskandari:
I thought I fixed that @adamralph ...I'm testing it again now.

HEskandari:
@adamralph should be good now. The latest commit now also clears the Search query and state properly.

src/ServiceInsight/Saga/SagaWindowViewModel.cs
```diff
@@ -99,6 +106,12 @@ string GetMajorVersion(StoredMessage message)
             return version?.Split('.').First();
         }
 
+        void ClearSaga()
+        {
+            Data = null;
+            NotifyOfPropertyChange(() => Data);
```
distantcam: This line is redundant. Setting the property will notify of the change anyway because of PropertyChanged.Fody.

src/ServiceInsight/Search/SearchBarViewModel.cs
```diff
@@ -109,6 +109,18 @@ public void SetupPaging(PagedResult<StoredMessage> pagedResult)
...
+        public void ClearPaging()
+        {
+            Result = new List<StoredMessage>();
+            CurrentPage = 0;
+            TotalItemCount = 0;
+            SearchQuery = null;
+            SearchInProgress = false;
+            SelectedEndpoint = null;
+
+            NotifyPropertiesChanged();
```
HEskandari: Okay, let me see if it is actually needed.

src/ServiceInsight/Saga/SagaWindowViewModel.cs
HEskandari: Agreed, I'll remove.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -245,6 +251,12 @@ void TryRebindMessageList(PagedResult<StoredMessage> pagedResult)
             }
         }
 
+        void ClearResult()
+        {
+            BindResult(new PagedResult<StoredMessage>());
+            SearchBar.ClearPaging();
+        }
+
```
adamralph: Is it worth creating this separate method instead of inlining in `Handle(ServiceControlConnectionChanged)`?

src/ServiceInsight/Saga/SagaWindowViewModel.cs
```diff
@@ -99,6 +106,11 @@ string GetMajorVersion(StoredMessage message)
             return version?.Split('.').First();
         }
 
+        void ClearSaga()
+        {
+            Data = null;
+        }
+
```
adamralph: Is it worth creating this separate method instead of inlining in `Handle(ServiceControlConnectionChanged)`?

HEskandari: Specially with this kind of code, it wouldn't be as clear why we're doing it when done directly int he event handler method.

src/ServiceInsight/MessageList/MessageListViewModel.cs
HEskandari: I think separate method would capture the intent better than inlining.

adamralph: Fair enough, perhaps it does indeed have some value. It may just be a matter of taste. I'm OK to have the separate method.

src/ServiceInsight/Saga/SagaWindowViewModel.cs
adamralph: Fair enough, perhaps it does indeed have some value. It may just be a matter of taste. I'm OK to have the separate method.

PR #629, opened 2016-11-14
Endpoint rewrite

distantcam:
This PR was done in stages

- Introduces RxServiceControl
- Fixes a threading issue with logging
- Rewrote the EndpointExplorerViewModel

### DIFF
--- a/src/ServiceInsight/App.xaml.cs
+++ b/src/ServiceInsight/App.xaml.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Diagnostics;
     using System.Windows;
+    using Akavache;
     using Anotar.Serilog;
     using DevExpress.Xpf.Core;
     using Framework.Logging;
@@ -40,6 +41,7 @@
     {
         public App()
         {
+            BlobCache.ApplicationName = "ServiceInsight";
             LoggingConfig.SetupLogging();
             if (!Debugger.IsAttached)
             {
--- a/src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
+++ b/src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
@@ -1,20 +1,33 @@
 ﻿namespace ServiceInsight.ServiceControl
 {
     using Caliburn.Micro;
+    using ServiceInsight.Framework;
 
     public class ServiceControlConnectionProvider
     {
         IEventAggregator eventAggregator;
+		IRxServiceControl serviceControl;
 
-        public ServiceControlConnectionProvider(IEventAggregator eventAggregator)
+        public ServiceControlConnectionProvider(IEventAggregator eventAggregator, IRxServiceControl serviceControl)
         {
             this.eventAggregator = eventAggregator;
+			this.serviceControl = serviceControl;
         }
 
         public void ConnectTo(string url)
         {
             Url = url;
-            eventAggregator.PublishOnUIThread(new ServiceControlConnectionChanged());
+			
+			eventAggregator.PublishOnUIThread(new ServiceControlConnectionChanged());
+
+            AsyncPump.Run(async () =>
+            {
+                await serviceControl.ClearServiceControls();
+                await serviceControl.AddServiceControl(url);
+
+                // Prime the streams
+                await serviceControl.Refresh();
+            });
         }
 
         public string Url { get; private set; }
--- a/src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
+++ b/src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
@@ -6,19 +6,17 @@
     public class ServiceControlConnectionProvider
     {
         IEventAggregator eventAggregator;
-		IRxServiceControl serviceControl;
+        IRxServiceControl serviceControl;
 
         public ServiceControlConnectionProvider(IEventAggregator eventAggregator, IRxServiceControl serviceControl)
         {
             this.eventAggregator = eventAggregator;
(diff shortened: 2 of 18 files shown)


HEskandari:
@distantcam Is this ready for review or a WIP?

distantcam:
@HEskandari It is ready for review, but I've noticed there's a conflict. I'll rebase this PR.

src/ServiceInsight/App.xaml.cs
```diff
@@ -40,6 +41,7 @@ public partial class App : IAppCommands
     {
         public App()
         {
+            BlobCache.ApplicationName = "ServiceInsight";
```
HEskandari: Does the cache has a schema? how does it cope with breaking changes? do we need to version it - by maybe by putting the version number in the cache name / storing folder? How do we guarantee backwards compatibility of the cache when we roll out a new version (since we're not strict on breaking changes).

src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
```diff
@@ -1,20 +1,33 @@
...
             this.eventAggregator = eventAggregator;
+			this.serviceControl = serviceControl;
         }
 
         public void ConnectTo(string url)
         {
             Url = url;
-            eventAggregator.PublishOnUIThread(new ServiceControlConnectionChanged());
+			
+			eventAggregator.PublishOnUIThread(new ServiceControlConnectionChanged());
```
HEskandari: Wouldn't this raise the ConnectionChanged event before actually changing the connection (below)?

src/ServiceInsight/Shell/ShellViewModel.cs
```diff
@@ -127,6 +139,24 @@ protected override void OnViewAttached(object view, object context)
...
         }
 
+        protected override void OnInitialize()
+        {
+            base.OnInitialize();
+
+            var configuredConnection = GetConfiguredAddress(comandLineArgParser);
+            var existingConnection = connectionProvider.Url;
+
+            eventAggregator.PublishOnUIThread(new WorkStarted("Trying to connect to ServiceControl"));
```
HEskandari: Shouldn't this be using the disposable WorkNotifer thingy?

src/ServiceInsight/Shell/ShellViewModel.cs
```diff
@@ -203,14 +233,16 @@ public void ConnectToServiceControl()
 
             if (result.GetValueOrDefault(false))
             {
-                EndpointExplorer.ConnectToService(connectionViewModel.ServiceUrl);
                 eventAggregator.PublishOnUIThread(new WorkFinished("Connected to ServiceControl Version {0}", connectionViewModel.Version));
```
HEskandari: Same as above?

src/ServiceInsight/App.xaml.cs
distantcam: All caching behavior is hit or miss, as in the cache hits, and you get a result, or it misses, and you have to do the extra work to create the value. For breaking changes, new versions, etc etc it just means the cache misses the first time, and then the new schema/value/whatever is stored in the cache.

src/ServiceInsight/Shell/ShellViewModel.cs
distantcam: #621 wasn't merged when I wrote this. I'll go fix that.

src/ServiceInsight/ServiceControl/ServiceControlConnectionProvider.cs
distantcam: Hmm, yes, but theoretically the things that use the new RxServiceControl won't need to listen to that event because they should get notified something changed reactively. I'll move the call anyway because yes it should change, then notify of change.

adamralph:
> This PR was done in stages

@distantcam I'm assuming the stages map to the commits as follows:

* "Introduces RxServiceControl" → commits 1, 2, 3
* "Fixes a threading issue with logging" → commit 4
* "Rewrote the EndpointExplorerViewModel" → commits 5, 6

distantcam:
@adamralph Pretty much. 2 is actually a bugfix for selection behaviour. The behaviour only worked one way. ViewModel updates when the View is changed, but if the ViewModel is changed the View doesn't update.

adamralph:
@distantcam when you say "bugfix" do you mean a fix for a bug which is observable by the user?

distantcam:
Part of me is tempted to cherry pick 2 and 4 out of this PR and apply them more immediately. The logging has always been fragile, which is why it breaks whenever I touch it. They also aren't really part of this PR except they are needed for this PR to work.

So, should I leave them in this PR or break them out?

@adamralph To answer your question, 2 is not currently not observable by the user. But every time I touch the logging code by adding another logger, or change the way logging is displayed then it does break. 4 might be observable, but only in the use-case when SI is opened from SP. I can verify that tomorrow if you'd like. Or you can check if you're up to it. You just need to Debug SI with a command line parameter. See https://docs.particular.net/serviceinsight/application-invocation

adamralph:
@distantcam 👍 to picking them out into a separate PR. I'd find it much easier to review those things in isolation.

I'll attempt the repro of the problem 4 is fixing, but I'm not 100% sure what I'm looking for.

distantcam:
No problem. I might close this PR then and redo it with less commits so it's cleaner in the end.

HEskandari:
One more thing that is not evident from the code reviews, is the usage of tabs, which makes the source code not compile. How does it work on your local visual studio @distantcam? (that's also the reason the build on TC is failing)

PR #630, opened 2016-11-14
Orders saga data before displaying them. Resolves #598.

### LINKED ISSUE #598: Saga View shows updates in incorrect order
image

I'm pretty sure that this is related to [change in ServiceControl that happened in April](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/SagaAudit/SagaDetailsIndex.cs#L52).

ServiceInsight is displaying changes in the order that SC serves them up. Is this a UI concern? Should we change it in SC or should SI always define a sort order?

HEskandari:


adamralph:
FYI I added "Fixes #598." to the description. That way we get a link to the PR from the issue.

PR #632, opened 2016-11-15
fix company name

adamralph:
AFAIK "Particular Software Incorporated" isn't a thing, and both SC and SP state "NServiceBus Ltd.". It's also the legal entity name used in the copyright.

I really didn't want to label this as a feature, but it does have external effect, so can't be a refactoring.


adamralph:
Added to 1.5.10 milestone.

PR #634, opened 2016-11-17
address code analysis warnings

adamralph:
I noticed these warnings while reviewing #629 

image

### DIFF
--- a/src/ServiceInsight.Tests/SearchViewModelTests.cs
+++ b/src/ServiceInsight.Tests/SearchViewModelTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace ServiceInsight.Tests
 {
+    using System;
     using NSubstitute;
     using NUnit.Framework;
     using ServiceInsight.Explorer.EndpointExplorer;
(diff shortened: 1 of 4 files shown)


src/ServiceInsight.Tests/SearchViewModelTests.cs
```diff
@@ -11,20 +12,24 @@
     using Shouldly;
 
     [TestFixture]
-    public class SearchViewModelTests
+    public sealed class SearchViewModelTests : IDisposable
```
adamralph: It's because they implement `IDisposable`. If we don't seal them then we have to implement the [entire pattern](https://msdn.microsoft.com/en-us/library/fs2xkftw.aspx) or otherwise yet another warning is given.

Generally, I agree regarding `sealed` classes and I'd be hesitant to seal anything in the production code, but these are just tests. If we ever need to inherit from them (highly unlikely) then we can unseal them and pay any further `IDisposable` tax at that time. We could also introduce Fody.Janitor instead but I think it's overkill for addressing these two violations.

distantcam: Yeah `IDisposable` is a dinosaur. Ok I'm happy to let this go, given it's just for a test anyway.

adamralph:
Added to 1.5.10 milestone.

Issue #637, opened 2016-11-18
Searching does not select the message or endpoint in the UI


distantcam:
When searching for a message the found message is then selected for diagrams, but is not selected in the message list. Also the endpoint for that message is not selected in the endpoint list.

This also affects links from ServicePulse, as they use the search system to select the message from ServicePulse.

HEskandari:
Does the endpoint has to get selected? If the root node (SC instance) is selected wouldn't that be enough? Also note that currently when you search by a message Id (just pasting the Id of the message in the search textbox), it will also bring up the messages that correlate to that message via RelatedTo field, which potentially could be from another endpoint.

distantcam:
@HEskandari The endpoint doesn't have to be selected, but currently even the root node is not selected. I didn't know about the RelatedTo bit.

The si: link I was testing with was when you right click on a message and get the link for it. That link also specifies the endpoint, which is selected if you use that link, but that selection is not shown in the UI.

HEskandari:
@distantcam Yes I remember search API tries to select the endpoint as well. Can't think of a scenario where that is really necessary though, since you're already filtering by message id and changing the selected endpoint in the Explorer is effectively a kind of filtering as well. 

We can 'try' to select the endpoint but think it wouldn't work in edge cases.

adamralph:
Note that the case of a search returning zero results also needs to be handled correctly.

HEskandari:
Just to clarify, the case where the search returns zero result, the views should be cleared. They all do, except Body tab.

We could also argue that it is a different bug, since it is about when nothing is selected which is the current behavior.

sergioc:
We could just pre-select the first msg in the search results.

kbaley:
Triaged with @boblangley. We've re-labeled this as an improvement rather than a bug.

PR #638, opened 2016-11-18
Internal fixes

distantcam:
Taken from #629 these are some small fixes.

- The log threading was brittle, but wasn't causing external issues
- Endpoint viewer selection behaviour was broken, and caused part of #637 
- Cleaned up some selection code


adamralph:
> Endpoint viewer selection behaviour was broken, and caused part of #637

Does this mean this PR is fixing one symptom described in #637?

distantcam:
@adamralph Yes. The endpoint fix came because I was focused on the endpoint list, and thus fixed that. Commenting on the old PR last night made me think to check searching which is where I discovered #637 which is a problem with message selection.

PR #640, opened 2016-11-22
Fixing ordering of handlers in SD

### LINKED ISSUE #570: Sequence Diagram misplaces Handlers

HEskandari:
~This would fix #570. This is a WIP.~

Fixes #570.

### DIFF
--- a/src/ServiceInsight.Tests/ServiceInsight.Tests.csproj
+++ b/src/ServiceInsight.Tests/ServiceInsight.Tests.csproj
@@ -137,6 +137,7 @@
     <Compile Include="SequenceDiagramModelCreatorTestsForJson1.cs" />
     <Compile Include="SequenceDiagramModelCreatorTestsForJson2.cs" />
     <Compile Include="SequenceDiagramModelCreatorTestsForJson3.cs" />
+    <Compile Include="SequenceDiagramModelCreatorTestsForJson5.cs" />
     <Compile Include="SequenceDiagramModelCreatorTestsForJson4.cs" />
     <Compile Include="ServiceControlConnectionDialogTests.cs" />
     <Compile Include="MessageListViewModelTests.cs" />
@@ -152,6 +153,7 @@
     <None Include="ConversationsData\json1.json" />
     <None Include="ConversationsData\json2.json" />
     <None Include="ConversationsData\json3.json" />
+    <None Include="ConversationsData\json5.json" />
     <None Include="ConversationsData\json4.json" />
     <None Include="packages.config" />
     <None Include="ServiceInsight.Tests.ruleset" />
--- a/src/ServiceInsight/SequenceDiagram/ModelCreator.cs
+++ b/src/ServiceInsight/SequenceDiagram/ModelCreator.cs
@@ -15,8 +15,8 @@ namespace ServiceInsight.SequenceDiagram
         readonly List<StoredMessage> messages;
         readonly IMessageCommandContainer container;
 
+        List<Handler> handlers;
         List<EndpointItem> endpoints = new List<EndpointItem>();
-        List<Handler> handlers = new List<Handler>();
         List<MessageProcessingRoute> processingRoutes = new List<MessageProcessingRoute>();
 
         public ModelCreator(List<StoredMessage> messages, IMessageCommandContainer container)
@@ -37,6 +37,7 @@ namespace ServiceInsight.SequenceDiagram
         {
             var endpointRegistry = new EndpointRegistry();
             var handlerRegistry = new HandlerRegistry();
+            var firstOrderHandlers = new List<Handler>();
 
             var messageTrees = CreateMessageTrees(messages).ToArray();
             var messagesInOrder = messageTrees.SelectMany(x => x.Walk()).ToArray();
@@ -71,7 +72,7 @@ namespace ServiceInsight.SequenceDiagram
 
                 if (handlerRegistry.TryRegisterHandler(CreateSendingHandler(message, sendingEndpoint), out sendingHandler))
                 {
-                    handlers.Add(sendingHandler);
+                    firstOrderHandlers.Add(sendingHandler);
                     sendingEndpoint.Handlers.Add(sendingHandler);
                 }
 
@@ -79,7 +80,7 @@ namespace ServiceInsight.SequenceDiagram
 
                 if (handlerRegistry.TryRegisterHandler(CreateProcessingHandler(message, processingEndpoint), out processingHandler))
                 {
-                    handlers.Add(processingHandler);
+                    firstOrderHandlers.Add(processingHandler);
                     processingEndpoint.Handlers.Add(processingHandler);
                 }
                 else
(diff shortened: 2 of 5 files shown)


HEskandari:

distantcam:
PR looks fine. I guess close the issue too?

src/ServiceInsight.Tests/ServiceInsight.Tests.csproj
```diff
@@ -167,8 +169,5 @@
       <Name>ServiceInsight</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup>
-    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
-  </ItemGroup>
```
adamralph: IIRC this speeds up Visual Studio's test discovery. Is there a reason it's been removed?

src/ServiceInsight/MessageViewers/XmlViewer/XmlMessageView.xaml.cs
```diff
@@ -1,6 +1,5 @@
 ﻿namespace ServiceInsight.MessageViewers.XmlViewer
 {
-    using System;
```
adamralph: I'm OK to merge as-is, but this really be in a separate commit.

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
```diff
@@ -4,6 +4,7 @@ namespace ServiceInsight.SequenceDiagram
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
     using System.Linq;
+    using DevExpress.Data.Filtering.Helpers;
```
adamralph: This has been added for no reason. It's unused.

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
```diff
@@ -25,6 +26,7 @@ public ModelCreator(List<StoredMessage> messages, IMessageCommandContainer conta
             this.container = container;
 
             Initialize();
+            OrderHandlers();
```
adamralph: I'm not really keen on the introduction of a second method which reassigns the `handlers` field. Can we encapsulate the ordering inside `Initialize()` and avoid the reassignment?

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
```diff
@@ -100,6 +102,17 @@ void Initialize()
             }
         }
 
+        private void OrderHandlers()
+        {
+            var start = handlers.Where(h => h.ID == ConversationStartHandlerName);
```
adamralph: Why do we have to special case the conversation start? Is this because `HandledAt` is not granular enough and the conversation start can have the same value as the next handler?

src/ServiceInsight.Tests/ServiceInsight.Tests.csproj
HEskandari: Not really. VS is just haphazardly removes or adds it. I can undo this change.

src/ServiceInsight/MessageViewers/XmlViewer/XmlMessageView.xaml.cs
HEskandari: I'll remove.

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
HEskandari: I'll remove.

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
HEskandari: this is about ordering to the best of our knowledge as the endpoints time could be out of sync. We know that logically the start of the conversation happens first, that's why it was special cased.

adamralph: OK, makes sense.

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
HEskandari: The main thing is that `Initialize` is doing a few things, including a first-step ordering. My thinking was that doing it as a part of Initialize (also a generalized name) would not making the second-step ordering obvious enough.

adamralph: What if `Initialize` created a list as a local variable and then assigned the `handlers` fields at the end, using the ordering? (The main code-smell is the reassignment of the `handlers` field.)

HEskandari: okay, I understand now. Let me give it another go.

HEskandari: @adamralph Please review again. Think I've addressed all the issues.

src/ServiceInsight/SequenceDiagram/ModelCreator.cs
```diff
@@ -98,6 +99,19 @@ void Initialize()
 
                 sendingHandler.Out = sendingHandler.Out.Concat(new[] { arrow }).OrderBy(a => a).ToList();
             }
+
+            ReOrderHandlers(firstOrderHandlers);
+        }
+
+        private void ReOrderHandlers(IList<Handler> orderedHandlers)
```
adamralph: This method has mixed responsibilities:

- order the handlers
- assign the `handlers` field

Can we remove this method and inline the code in `Initialize`? It's only 4 statements, so I'm having a hard time justifying its existence.

distantcam: I disagree with this assessment.

The method is badly named, and could take an `IEnumerable` instead of an `IList`, which would make it more clear that it's assigning the handlers field with the values from the parameter. I don't think that's a mixed responsibility.

I'm also against inlining it. `Initialize` is long enough. It makes the code easier to understand, so there's the justification for it's existence.

HEskandari: I was thinking the same way @distantcam. I think maybe `Initialize` is not a good name, and we should split that, not adding to the problem. It is already too large to comprehend.

HEskandari: Let's discuss on our sync tomorrow.

HEskandari: @adamralph @distantcam as discussed, for now I have merged the ordering method with initialize.

distantcam:
@adamralph @HEskandari I thought the plan was to refactor the initialise method in another commit before this PR was ready. Is that not the case?

HEskandari:
@distantcam My understanding was that we should do it separate as a Refactoring PR. I'll submit that separately.

distantcam:
I'm -1 for a separate PR. I'm +1 for another commit to this PR.

adamralph:
IIRC, when we discussed, this I did indeed suggest "separate refactoring commit" rather than "separate refactoring PR".

After reviewing the new diff, I can see that the this change is relatively trivial (a few extra lines in the method) so in this case I'd prefer that we merge the PR as-is (after a squash) so that the bug is fixed and we can release the patch. The refactoring is likely to have a significantly more complex diff, and I expect we'll have a few opinions on it, so I'd rather deal with that asynchronously.

distantcam:
> The refactoring is likely to have a significantly more complex diff, and I expect we'll have a few opinions on it, so I'd rather deal with that asynchronously.

@adamralph Fair enough. Did you want to approve this PR then?

HEskandari:
I'll squash the commits and update the PR.

HEskandari:
@Particular/serviceinsight-maintainers The commits are merged. Please have one last review and merge if it's okay.

PR #646, opened 2016-11-23
Log threading fix

distantcam:
This fixes up the threading in the logging system.

Lesson's learned:
- ObserveOn is almost always the right thing, SubscribeOn is almost always the wrong thing
- `Do` needs the observer scheduler set _before_ `Do` is called

### DIFF
--- a/src/ServiceInsight/Framework/Logging/LoggingConfig.cs
+++ b/src/ServiceInsight/Framework/Logging/LoggingConfig.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.IO;
+    using System.Reactive.Concurrency;
     using System.Reactive.Linq;
     using Caliburn.Micro;
     using LogWindow;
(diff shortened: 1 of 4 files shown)


HEskandari:
In which condition would the threading issue happen? How can I test before/after and make sure it is fixed?

distantcam:
Take out the filter in the logging config

```
.WriteTo.Logger(lc => lc
    .MinimumLevel.Verbose()
    .Filter.ByIncludingOnly(Matching.FromSource<IServiceControl>()) // Comment out this line
    .WriteTo.Observers(logEvents => logEvents
        .ObserveOn(TaskPoolScheduler.Default)
        .Do(LogWindowViewModel.LogObserver)
        .Subscribe()))
```

Then add a log somewhere it will be triggered, but make sure it runs on a new thread.

    new Thread(() => { LogTo.Information("Boom"); }).Start();

I recommend adding that line to the top of `T Execute<T, T2>(RestRequestWithCache request, Func<IRestResponse<T2>, T> selector)` in `DefaultServiceControl` but you could add it pretty much anywhere.

The app should crash on startup.

HEskandari:
@distantcam looks good. Merging.

adamralph:
Added to 1.5.10 milestone.

Issue #641, opened 2016-11-22
Using the 'Detail view' to show the all message payloads in Saga View


HEskandari:
- Showing the 'Detail view' for the all objects (in the middle panel and right / left hand panels) as a generic way to display saga data. 
- The threshold and use of ellipsis when the content gets long should be removed.
- User will be able to see all the trimmed property names and values in the detail view.

This is a split of #624.

adamralph:
This is labelled as a bug, but I'm finding it hard to parse the description to know what the symptom of the bug is, and how it should be fixed. Can we improve the description somehow?

HEskandari:
@adamralph I no longer think this should be labelled as a bug. It is an improvement on how the message payloads are displayed. Currently they are displayed in expandable panels, all loaded at once when clicking the Show Message button.  I'll update the description and add what I think should be the expected behavior.

adamralph:
Thanks @HEskandari. I made another tweak to the title so it reads more like a feature (e.g. in release notes).

HEskandari:
Reviewed. This is still the case, but requires substantial work to redesign the view.

PR #648, opened 2016-11-24
[WIP] Displaying <Null> for empty/null property values in SagaView. 

### LINKED ISSUE #642: Null values are not represented in SagaView
In the SagaView panels the null values are left empty on right hand side of the equal signs ("="). The lack of value should be clearly communicated. 

Options are using `<Null>`, `<Nil>` or `N/A`.

This is a split of #624.

Depends on Particular/ServiceControl#889

HEskandari:
Resolves #642

The saga data values are are stored as serialized json objects without types information. All the primitive values are received in SI as a string. The lack of value were signified with an empty string, which is now displayed as "<Null>" instead. Given the lack of type information, I think this is the best that could be done.

Here's how it looks now:

image

@Particular/serviceinsight-maintainers please review at your convenience.

### DIFF
--- a/src/ServiceInsight/Saga/SagaUpdatedValue.cs
+++ b/src/ServiceInsight/Saga/SagaUpdatedValue.cs
@@ -7,12 +7,13 @@
     public class SagaUpdatedValue : PropertyChangedBase
     {
         public const byte MaxValueLength = 30;
+        public const string NullValue = "null";
 
         public SagaUpdatedValue(string sagaName, string propertyName, string propertyValue)
         {
             SagaName = sagaName;
             Name = propertyName;
-            NewValue = propertyValue;
+            NewValue = propertyValue ?? NullValue;
             ShowEntireContentCommand = this.CreateCommand(ShowEntireContent);
         }
 
@@ -53,7 +54,7 @@
 
         public void UpdateOldValue(SagaUpdatedValue oldValueHolder)
         {
-            OldValue = oldValueHolder != null ? oldValueHolder.NewValue : string.Empty;
+            OldValue = oldValueHolder != null ? oldValueHolder.NewValue ?? NullValue : string.Empty;
         }
     }
 }


sergioc:
This is not a string, right? Why not just display `Null` without any kind of additional characters around it?

adamralph:
@sergioc the problem there is that the value itself might be the string "null". How do we display that?

Given that all the values that SI receives from SC are strings, the best we can do is to wrap them in double quotes, leaving `null` as a non-string value, communicating "nullness".

Or is there another way you can suggest?

HEskandari:
Should we threat it like that across the board? Or should it be special-cased for 'simple' values? e.g.. consider these cases:

- When the value is a boolean string "True" or "False"
- When the value is empty collection: "[]"
- When the value is an object graph: `"'PrimaryContact':{'PersonId':'1'}"` (a simplified version, could be multiple/nested objects.

My personal take is that it looks a bit weird. Why would you warp "[]" in a quotation where on its own it is a meaningful concept.

sergioc:
Unless it's an actual string, I suggest not wrapping in quotes or any other kind of characters. E.g. the words for boolean True or False are technically strings, but they don't represent strings. So no wrap for that.

> the problem there is that the value itself might be the string "null". How do we display that?

if the value itself is that, then show `"null"`. Else we can be explicit and say `no data` or `empty property` or something custom like that. A custom status probably has a higher chance of not being mistaken by an actual `null` string.

HEskandari:
@sergioc That's the issue! We have no way of knowing if it is a string, or a boolean wrapped in a string. Is the string null? or is it an object that doesn't exist.

With the lack of type information, there's no way to distinct any of these.

adamralph:
Whichever text is suggested as being an indicator for non-existent value, it will always have the problem of clashing with a string value which contains that same text, so there must be some kind of visual distinction given to the text which means "no value". Perhaps, instead of doing this via content (e.g. quotation marks) we can do this with style, e.g. the text `null` but greyed out, or in italics, etc.?

sergioc:
> greyed out

I like this suggestion, assuming we increase contrast for the currently used color.

HEskandari:
@sergioc would you be able to share you thoughts via a screenshot, to make sure I got the style/colors correctly?

sergioc:
@HEskandari yep. I'll get back to you early next week.

distantcam:
@sergioc Any luck taking a look at this?

sergioc:
@distantcam I should have some time today.

sergioc:
Here's a general improvement to the saga properties display:

2016-12-01 at 19 10

I increased contrast for property names and values while keeping special characters de-emphasized. I also show the active tab in underlined black text to show it's active and can't be clicked again.

Details:

2016-12-01 at 19 07

HEskandari:
Here's how it looks now. I've fixed the cursor for the 'link' texts as well. The only thing I was reluctant to change was the new way of displaying the 'changed bytes'. The old (current) implementation shows as "xxxx byte(s) -> yyyy byte(s)" but the screenshot above shows them as "(xxxx bytes) -> (yyyy bytes)". Since there is issue #641 to change that behavior, don't think it is necessary to change it now.

image

adamralph:
I like the changes, but I would have preferred to keep this PR scoped to just fixing #642.

This PR now introduces other improvements, and the release notes for the next version should reflect them. I suggest we raise a feature issue titled something like "Improved saga data view", mark it as a feature and also link this PR to it. The version in which that feature is released should then be a minor and not a patch.

Thoughts?

src/ServiceInsight/Saga/SagaResources.xaml
```diff
@@ -11,6 +13,26 @@
...
+        <Setter Property="Foreground" Value="{StaticResource SagaBlue}" />
+        <Setter Property="Cursor" Value="Hand" />
+        <Style.Triggers>
+            <Trigger Property="IsMouseOver" Value="True">
+                <Setter Property="TextDecorations" Value="Underline" />
+            </Trigger>
+        </Style.Triggers>
+    </Style>
+    
+    <Style x:Key="NewValueDisplay" TargetType="{x:Type TextBlock}">
```
distantcam: I don't like this template. It works on a "magic string" value, which was originally the string we were going to show for null values.

Instead, I'd prefer the VM have a bool, and the textbox showing the null is shown/hidden based on that bool.

distantcam:
I agree with @adamralph about this PR. It's fixed the `null` value, and it looks good now with the colour change and italics. Other changes can be discussed and implemented in another issue.

HEskandari:
@adamralph @distantcam So you want everything to be wrapped in a "" then? as I said, there is no type and *everything* is a string.

HEskandari:
@adamralph I agree on principle but given we're here seems the most straight way forward would be to create the issue you mentioned.

HEskandari:
We've decided to keep the scope of this PR to fix the bug only. The styling part will be done separately as a part of issue #666. I'll update the PR accordingly.

HEskandari:
@adamralph I tried to reproduce a sample where we display a "null" value on the screen (and then apply the styling) but with the latest proposed and applied changes it is not possible anymore. All the null values come through as an empty string. Here's a property of type _DateTime?_ and you can see from the screenshot how it is displayed. The code screenshot shows the actual values.

image

image

adamralph:
@HEskandari let's jump on zoom again or chat in Lisbon about this one.

HEskandari:
Closing this for now as it is not possible to do this with the current state of ServiceControl.

PR #650, opened 2016-11-29
Removing binary utf8 preamble. 

### LINKED ISSUE #649: Body tab removes the first character when payload is Xml
Steps to reproduce:

- Have an endpoint that uses XML serialization.
- Generate messages that goes through SC
- Connect the SI and switch to the the body view for a message
- First letter of the content is removed in SI, e.g. "<xml " becomes "xml"

ServiceControl sends the data correctly to ServiceInsight. This is due to how the UTF preamble is removed. It seems removing it as "string" would not work properly and the invisible strings should be removed from the byte array instead.

HEskandari:
Resolves #649

### DIFF
--- a/src/ServiceInsight/ExtensionMethods/BytesExtensions.cs
+++ b/src/ServiceInsight/ExtensionMethods/BytesExtensions.cs
@@ -1,0 +1,13 @@
+﻿namespace ServiceInsight.ExtensionMethods
+{
+    using System.Linq;
+
+    public static class BytesExtensions
+    {
+        public static bool StartsWith(this byte[] source, byte[] comparison) =>
+            source != null &&
+            comparison != null &&
+            source.Length >= comparison.Length &&
+            source.Take(comparison.Length).SequenceEqual(comparison);
+    }
+}
--- a/src/ServiceInsight/Saga/JsonPropertiesHelper.cs
+++ b/src/ServiceInsight/Saga/JsonPropertiesHelper.cs
@@ -1,6 +1,5 @@
 ﻿namespace ServiceInsight.Saga
 {
-    using System;
     using System.Collections.Generic;
     using System.Linq;
     using Newtonsoft.Json;
@@ -9,11 +8,11 @@
     {
         static readonly IList<string> StandardKeys = new List<string> { "$type", "Id", "Originator", "OriginalMessageId" };
 
-        public static IList<KeyValuePair<string, string>> ProcessValues(string stateAfterChange, Func<string, string> cleanup) => ProcessValues(cleanup(stateAfterChange));
-
-        static IList<KeyValuePair<string, string>> ProcessValues(string stateAfterChange) => JsonConvert.DeserializeObject<Dictionary<string, object>>(stateAfterChange)
+        public static IList<KeyValuePair<string, string>> ProcessValues(string stateAfterChange) => JsonConvert.DeserializeObject<Dictionary<string, object>>(stateAfterChange)
                   .Where(m => StandardKeys.All(s => s != m.Key))
                   .Select(f => new KeyValuePair<string, string>(f.Key, f.Value == null ? string.Empty : f.Value.ToString()))
                   .ToList();
+
+        public static IList<KeyValuePair<string, string>> ProcessArray(string stateAfterChange) => ProcessValues(stateAfterChange.TrimStart('[').TrimEnd(']'));
     }
 }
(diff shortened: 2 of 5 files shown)


src/ServiceInsight/ExtensionMethods/BytesExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+        {
+            if (source != null && comparison != null &&
+                source.Length >= comparison.Length &&
+                source.Take(comparison.Length).SequenceEqual(comparison))
+            {
+                return true;
+            }
+
+            return false;
+        }
```
adamralph: This is just a style comment, but this method could be:

```C#
public static bool StartsWith(this byte[] source, byte[] comparison) =>
    source != null &&
    comparison != null &&
    source.Length >= comparison.Length &&
    source.Take(comparison.Length).SequenceEqual(comparison);
```

src/ServiceInsight/Saga/JsonPropertiesHelper.cs
```diff
@@ -9,11 +8,14 @@ class JsonPropertiesHelper
...
-        static IList<KeyValuePair<string, string>> ProcessValues(string stateAfterChange) => JsonConvert.DeserializeObject<Dictionary<string, object>>(stateAfterChange)
+        public static IList<KeyValuePair<string, string>> ProcessValues(string stateAfterChange) => JsonConvert.DeserializeObject<Dictionary<string, object>>(stateAfterChange)
                   .Where(m => StandardKeys.All(s => s != m.Key))
                   .Select(f => new KeyValuePair<string, string>(f.Key, f.Value == null ? string.Empty : f.Value.ToString()))
                   .ToList();
+
+        public static IList<KeyValuePair<string, string>> ProcessArray(string stateAfterChange)
+        {
+            return ProcessValues(stateAfterChange.TrimStart('[').TrimEnd(']'));
+        }
```
adamralph: (style) could be:

```C#
public static IList<KeyValuePair<string, string>> ProcessArray(string stateAfterChange) =>
    ProcessValues(stateAfterChange.TrimStart('[').TrimEnd(']'));
```

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -480,7 +483,17 @@ T ProcessResponse<T>(Func<IRestResponse, T> selector, IRestResponse response)
             return new List<KeyValuePair<string, string>>();
         }
 
-        static string CleanupBodyString(string bodyString) => bodyString.StartsWith(byteOrderMarkUtf8) ? bodyString.Remove(0, byteOrderMarkUtf8.Length) : bodyString;
+        static void CleanResponse(IRestResponse response)
+        {
+            if (response.RawBytes.StartsWith(byteOrderMarkUtf8))
+            {
+                var strippedBytes = response.RawBytes.Skip(byteOrderMarkUtf8.Length)
+                                            .Take(response.RawBytes.Length - byteOrderMarkUtf8.Length)
```
adamralph: The call to `Take()` is redundant. The `Skip()` will skip the BOM and then the enumerator will continue until the end of the byte array.

HEskandari:
@adamralph Think the latest push addresses all your comments.

distantcam:
LGTM

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -480,7 +483,17 @@ T ProcessResponse<T>(Func<IRestResponse, T> selector, IRestResponse response)
             return new List<KeyValuePair<string, string>>();
         }
 
-        static string CleanupBodyString(string bodyString) => bodyString.StartsWith(byteOrderMarkUtf8) ? bodyString.Remove(0, byteOrderMarkUtf8.Length) : bodyString;
+        static void CleanResponse(IRestResponse response)
+        {
+            if (response.RawBytes.StartsWith(byteOrderMarkUtf8))
+            {
+                var strippedBytes = response.RawBytes
+                                            .Take(response.RawBytes.Length - byteOrderMarkUtf8.Length)
```
adamralph: @HEskandari your second commit removed the wrong call here. The `Skip()` should have remained. The `Take()` should have been removed.

HEskandari:
@adamralph d'oh! Pushed an update.

HEskandari:
@adamralph no objections.

HEskandari:
@adamralph Done.

PR #657, opened 2016-11-30
Added some graphical diff tests for XamlIcons.xaml

distantcam:


HEskandari:
There are some empty icons there, is that expected?

distantcam:
@HEskandari No it's not. So that's the next step, figure out why they're blank, and if they're used.

HEskandari:
@distantcam I've run the tests locally and they all pass. I wonder if this has something with Aero being disabled on server OS environments.

distantcam:
@HEskandari You're running Win10 Pro? I'm running 10.0.14393

HEskandari:
@distantcam I'm running Win10 Home - Version	10.0.14393 Build 14393

distantcam:
@HEskandari Same build as me, just a different key. Apparently the test changes depending on the OS version, which is why my approvals work on your machine.

HEskandari:
Yeah...I take that back...with Aero disabled they still pass here.

distantcam:
Ok, I've decided to mark these tests as explicit, since they are still worth having but cannot currently run on the server.

I've also fixed 3 of the broken icons. The test was not running enough of the WPF system for them to render properly.

This leaves 2 icons still broken (blank). Looking into those they're part of the SagaView that's never shown. There's a lot of dead code there so I'll remove that as part of another PR.

src/ServiceInsight.Tests/XamlIconsTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+
+        [Test]
+        [Explicit]
+        [TestCaseSource(typeof(XamlIconsTests), nameof(GetGeometries))]
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public void GeometriesTest(string resourceName)
+        {
+            var icon = new Path { Fill = Brushes.Black, Stretch = Stretch.UniformToFill };
```
adamralph: With `Stretch = Stretch.UniformToFill` the image looks like this, with the circle cropped:

image

but with `Stretch = Stretch.Uniform`, it looks like this, with the entire circle shown:

image

Would it be better to have the latter so that the full geometry is shown?

adamralph:
One more thing: the `ReceivingEndpoint` and `SendingEndpoint` control templates are being rendered as transparent images with no visible artifacts. Is that expected?

src/ServiceInsight.Tests/XamlIconsTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        public static IEnumerable GetDrawingImages => GetTestCases<DrawingImage>();
+
+        public static IEnumerable GetGeometries => GetTestCases<Geometry>();
+
+        [Test]
+        [Explicit]
+        [TestCaseSource(typeof(XamlIconsTests), nameof(GetControlTemplates))]
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public void ControlTemplateTest(string resourceName)
```
distantcam: I like this idea. Will change.

src/ServiceInsight.Tests/XamlIconsTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            var icon = new ContentControl { Foreground = Brushes.Green };
```
distantcam: It's not a color we use in the coloured icons. It's just for contrast. Any colour other than white can be used here.

src/ServiceInsight.Tests/XamlIconsTests.cs
distantcam: Yes it would be better, but I would consider that a separate PR. There might be side effects to fixing the icon without knowing where it's used.

src/ServiceInsight.Tests/XamlIconsTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            NamerFactory.Clear();
+            using (ApprovalResults.ForScenario(resourceName))
+            {
+                Approvals.Verify(new ImageWriter(f => WpfUtils.ScreenCapture(control, f)));
+            }
+        }
+
+        [Test]
+        public void OnlyExpectedResourcesInDictionary()
```
distantcam: The icon file is a resource dictionary, which can contain any type of resource, not just the 3 we are testing for (ControlTemplate, DrawingImage, Geometry). This test ensures that only resources of those 3 types are in the dictionary. That way if we add another way to draw icons, this test will fail and we know to update the icon tests.

distantcam:
> One more thing: the ReceivingEndpoint and SendingEndpoint control templates are being rendered as transparent images with no visible artifacts. Is that expected?

It's not expected exactly. This is related to #667. Those icons effectively aren't being used so in #667 I removed them and the code surrounding that.

distantcam:
@adamralph What version of windows are you running? The test pasts on mine and @HEskandari machines, and we're both running 10.0.14393.

src/ServiceInsight.Tests/XamlIconsTests.cs
distantcam: Wait, I just realized that's a test change, not an icon one. I need more caffeine. I'll make this change too.

distantcam:
> One thing that needs to be added is a line in gitignore to ignore the received files.

I don't like this idea. Having a received file means a) the test is still failing, or b) the test hasn't been run since it's been fixed. Either way signals an error. Once the test has been run and passes ApprovalTests will remove the file. So having the file turn up in the git changes means something is wrong.

> I found that UseApprovalSubdirectory can be applied at the method level like so:

I don't like this either. I do see your problem (especially if you're not getting a diff opening, what diff tool do you use?) but I think there's other ways to solve that problem.

adamralph:
@distantcam this is my Windows version:

adamralph:
> what diff tool do you use?

When I run the API approval tests in other projects, the diff shows in VS like so:

image

The only other diff viewer I have installed is the one in TortoiseGit.

adamralph:
> Having a received file means a) the test is still failing, or b) the test hasn't been run since it's been fixed. Either way signals an error. Once the test has been run and passes ApprovalTests will remove the file. So having the file turn up in the git changes means something is wrong.

Unless we can guarantee that the tests will pass in all environments, having received files doesn't necessarily mean there is anything wrong. E.g. on my system, having them on disk is currently business as normal.

distantcam:
@adamralph Can this PR be merged?

adamralph:
@distantcam I push this one further back in my to do list in favour of #661 but now that that one is merged, I'll take another look at this one.

src/ServiceInsight.Tests/AutoApprover.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace ServiceInsight.Tests
+{
+    using System.IO;
+    using ApprovalTests.Core;
+
+    class AutoApprover : IReporterWithApprovalPower
```
adamralph: Is there any value in keeping this class? We're not using it anywhere, since the corresponding attribute is commented out in `XamlIconsTests`.

adamralph: I guess this was fixed, so I don't know why the comment still shows. `¯\_(ツ)_/¯`

distantcam: Yes I want to keep this class. It's an easy way to approve all the tests, especially when they've changed.

distantcam:
@adamralph Regarding `.gitignore` and received files. I can see your point, given the tests "fail" on your system. I don't like the idea of ignoring _all_ received files. How about just ignoring the received files in the XamlIcons directory?

src/ServiceInsight.Tests/AutoApprover.cs
adamralph: OK, so it's a dev tool. That's fine. 👍

adamralph:
> How about just ignoring the received files in the XamlIcons directory?

Sure, that's good enough for me.

adamralph:
👍 thanks @distantcam

PR #660, opened 2016-12-02
Removed redundant code from MessageList

distantcam:
I was reviewing the MessageList selection code, and removed some unnecessary code.


adamralph:
BTW - I smoke tested it and everything looks fine.

PR #663, opened 2016-12-06
Code cleanup

HEskandari:
This cleans up a few code issues.

- Cleaning up two-way binding to getter-only properties
- Cleaning up usings
- Removing unused dependencies
- Fixing design-time Resharper and VisualStudio warnings

Each change is a separate commit to allow easier reviews. I can squash them into one after the review.
Please review @Particular/serviceinsight-maintainers

### DIFF
--- a/src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
+++ b/src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
@@ -39,7 +39,7 @@
             xmlContent = Substitute.For<XmlMessageViewModel>();
             selection = new MessageSelectionContext(eventAggregator);
 
-            messageBodyFunc = () => new MessageBodyViewModel(hexContent, jsonContent, xmlContent, serviceControl, eventAggregator, workNotifier, selection);
+            messageBodyFunc = () => new MessageBodyViewModel(hexContent, jsonContent, xmlContent, serviceControl, workNotifier, selection);
         }
 
         [Test]
--- a/src/ServiceInsight.Tests/packages.config
+++ b/src/ServiceInsight.Tests/packages.config
@@ -5,6 +5,7 @@
   <package id="Caliburn.Micro.Core" version="2.0.2" targetFramework="net45" />
   <package id="NSubstitute" version="1.10.0.0" targetFramework="net45" />
   <package id="NUnit" version="3.2.1" targetFramework="net45" />
+  <package id="ObservablePropertyChanged" version="0.1.3" targetFramework="net452" developmentDependency="true" />
   <package id="reactiveui-core" version="5.5.1" targetFramework="net45" />
   <package id="reactiveui-testing" version="5.5.1" targetFramework="net45" />
   <package id="RestSharp" version="105.0.1" targetFramework="net45" />
--- a/src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
+++ b/src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
@@ -17,7 +17,7 @@
                     <SolidColorBrush x:Key="GlyphBrush" Color="#444" />
                     <SolidColorBrush x:Key="DisabledForegroundBrush" Color="#888" />
                     <SolidColorBrush x:Key="WindowBackgroundBrush" Color="#000" />
-                    <SolidColorBrush x:Key="SelectedBackgroundBrush" Color="#007AC7" />
+                    <SolidColorBrush x:Key="SelectedNodeBackgroundBrush" Color="#007AC7" />
 
                     <LinearGradientBrush x:Key="NormalBrush" StartPoint="0,0" EndPoint="0,1">
                         <GradientBrush.GradientStops>
@@ -150,7 +150,7 @@
                                 <Setter TargetName="Icon" Property="Visibility" Value="Hidden" />
                             </Trigger>
                             <Trigger Property="IsHighlighted" Value="true">
-                                <Setter TargetName="Border" Property="Background" Value="{StaticResource SelectedBackgroundBrush}" />
+                                <Setter TargetName="Border" Property="Background" Value="{StaticResource SelectedNodeBackgroundBrush}" />
                             </Trigger>
                             <Trigger Property="IsEnabled" Value="false">
                                 <Setter Property="Foreground" Value="{StaticResource DisabledForegroundBrush}" />
@@ -210,7 +210,7 @@
                                 <Setter TargetName="Icon" Property="Visibility" Value="Collapsed" />
                             </Trigger>
                             <Trigger Property="IsHighlighted" Value="true">
-                                <Setter TargetName="Border" Property="Background" Value="{StaticResource SelectedBackgroundBrush}" />
+                                <Setter TargetName="Border" Property="Background" Value="{StaticResource SelectedNodeBackgroundBrush}" />
                             </Trigger>
                             <Trigger SourceName="Popup" Property="Popup.AllowsTransparency" Value="True">
                                 <Setter TargetName="SubmenuBorder" Property="CornerRadius" Value="4" />
--- a/src/ServiceInsight/MessageHeaders/MessageHeadersView.xaml
+++ b/src/ServiceInsight/MessageHeaders/MessageHeadersView.xaml
@@ -94,22 +94,22 @@
                 </Trigger>
             </Style.Triggers>
         </Style>
+        <DataTemplate x:Key="TextTemplate">
+            <TextBlock Text="{Binding Value}" d:DataContext="{d:DesignInstance messageHeaders:MessageHeadersKeyValueList}" />
+        </DataTemplate>
+        
+        <DataTemplate x:Key="ExceptionTemplate">
+            <controls:MvvmTextEditor d:DataContext="{d:DesignInstance messageHeaders:MessageHeadersKeyValueList}"
+                                     FontFamily="Consolas"
+                                     FontSize="10pt"
+                                     IsReadOnly="true"
+                                     ScrollViewer.CanContentScroll="False"
+                                     Style="{StaticResource MvvmTextEditorScrollFixStyle}"
+                                     SyntaxHighlighting="StackTrace"
+                                     Text="{Binding Value}" />
+        </DataTemplate>
     </UserControl.Resources>
     <Grid>
-        <Grid.Resources>
-            <DataTemplate x:Key="TextTemplate">
-                <TextBlock Text="{Binding Value}" />
-            </DataTemplate>
-            <DataTemplate x:Key="ExceptionTemplate">
-                <controls:MvvmTextEditor FontFamily="Consolas"
-                                         FontSize="10pt"
-                                         IsReadOnly="true"
-                                         ScrollViewer.CanContentScroll="False"
-                                         Style="{StaticResource MvvmTextEditorScrollFixStyle}"
-                                         SyntaxHighlighting="StackTrace"
-                                         Text="{Binding Value}" />
-            </DataTemplate>
-        </Grid.Resources>
         <DataGrid Name="gridView"
                   AutoGenerateColumns="False"
                   CanUserAddRows="False"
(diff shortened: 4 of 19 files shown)


adamralph:
I guess this probably needs to go in release notes as some kind of functional change?

HEskandari:
Maybe we need to highlight it a bit more in the announcement rather than an title in the release notes amongst a few other things that have changed?

adamralph:
Marking as WIP. As discussed, we'll leave the net452 upgrade aside for now and this PR will be reduced to just the refactorings.

HEskandari:
@Particular/serviceinsight-maintainers I have removed the commit related to 4.5.2 upgrade. This is ready for a review.

src/ServiceInsight/Shell/About.xaml
```diff
@@ -48,14 +48,14 @@
...
                 <TextBlock Style="{StaticResource InfoText}">
                     <Run Text="Version  " />
-                    <Run Text="{Binding AppVersion}" />
+                    <Run Text="{Binding AppVersion, Mode=OneWay}" />
                     <Run Text="  Commit hash:  " />
-                    <Run Text="{Binding CommitHash}" />
+                    <Run Text="{Binding CommitHash, Mode=OneWay}" />
                 </TextBlock>
 
                 <TextBlock Style="{StaticResource InfoText}" Visibility="{Binding IsSplash, Converter={StaticResource BoolToVisibilityConverterInverted}}">
```
adamralph: `IsSplash` is also read-only but this binding has not been changed to `OneWay`. Is this because the binding is being used for `Visibility` rather than `Text`? (Same question for `ShellViewModel.StatusBarManager.ErrorMessageVisible`.)

I also noticed that `CopyrightText` is also read-only but the binding has not been changed to `OneWay`. Is this because `CopyrightText` is bound to a `TextBlock` instead of a `Run`?

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
```diff
@@ -31,15 +30,14 @@ public class MessageBodyViewModelTests
...
             workNotifier = Substitute.For<IWorkNotifier>();
             serviceControl = Substitute.For<IServiceControl>();
             hexContent = Substitute.For<HexContentViewModel>();
             jsonContent = Substitute.For<JsonMessageViewModel>();
             xmlContent = Substitute.For<XmlMessageViewModel>();
-            selection = new MessageSelectionContext(eventAggregator);
+            selection = new MessageSelectionContext(Substitute.For<IEventAggregator>());
 
-            messageBodyFunc = () => new MessageBodyViewModel(hexContent, jsonContent, xmlContent, serviceControl, eventAggregator, workNotifier, selection);
+            messageBodyFunc = () => new MessageBodyViewModel(hexContent, jsonContent, xmlContent, serviceControl, workNotifier, selection);
```
adamralph: This change was included in the "Cleaning up usings" commit, but the related change to `MessageBodyViewModel` was not, so that commit does not compile.

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
```diff
@@ -2,24 +2,23 @@
...
+    using ServiceControl;
     using ServiceInsight.Framework;
     using ServiceInsight.Framework.Events;
-    using ServiceInsight.MessageList;
-    using ServiceInsight.MessageViewers;
-    using ServiceInsight.MessageViewers.HexViewer;
-    using ServiceInsight.MessageViewers.JsonViewer;
-    using ServiceInsight.MessageViewers.XmlViewer;
-    using ServiceInsight.Models;
-    using ServiceInsight.ServiceControl;
```
adamralph: Can this convention be enforced somehow?

Personally I much prefer the namespaces to be declared in full, but if the agreed convention is to shorten them then I won't argue against it here.

src/ServiceInsight/MessageHeaders/MessageHeadersView.xaml
```diff
@@ -94,22 +94,22 @@
...
-            <DataTemplate x:Key="ExceptionTemplate">
-                <controls:MvvmTextEditor FontFamily="Consolas"
-                                         FontSize="10pt"
-                                         IsReadOnly="true"
-                                         ScrollViewer.CanContentScroll="False"
-                                         Style="{StaticResource MvvmTextEditorScrollFixStyle}"
-                                         SyntaxHighlighting="StackTrace"
-                                         Text="{Binding Value}" />
-            </DataTemplate>
-        </Grid.Resources>
```
adamralph: I'm not really sure what this change is doing. @distantcam can you review this change please?

src/ServiceInsight.Tests/packages.config
```diff
@@ -5,6 +5,7 @@
   <package id="Caliburn.Micro.Core" version="2.0.2" targetFramework="net45" />
   <package id="NSubstitute" version="1.10.0.0" targetFramework="net45" />
   <package id="NUnit" version="3.2.1" targetFramework="net45" />
+  <package id="ObservablePropertyChanged" version="0.1.3" targetFramework="net452" developmentDependency="true" />
```
adamralph: Is this effectively a transitive dependency? I.e. if we stopped using `ObservablePropertyChanged` in the `ServiceInsight` project, would we still use it in `ServiceInsight.Tests` for something else? If the answer is no, I think it's probably better to leave it as transitive and omit it from this packages.config.

src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
```diff
@@ -17,7 +17,7 @@
                     <SolidColorBrush x:Key="GlyphBrush" Color="#444" />
                     <SolidColorBrush x:Key="DisabledForegroundBrush" Color="#888" />
                     <SolidColorBrush x:Key="WindowBackgroundBrush" Color="#000" />
-                    <SolidColorBrush x:Key="SelectedBackgroundBrush" Color="#007AC7" />
+                    <SolidColorBrush x:Key="SelectedNodeBackgroundBrush" Color="#007AC7" />
```
adamralph: What is the conflict that this is solving?

src/ServiceInsight/Shell/About.xaml
HEskandari: So the default value is "Default" which means it uses the value from the dependency property which varies between the controls. So some controls default to one-way (usually read-only controls like TextBlock) and some use two-way binding (like editable controls such as TextBox). 

What is a strange default in fact, is that Run.Text property uses Two-Way as default and that *could* be because it is used in other controls such as RichTextBox which supports two-way binding.

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
HEskandari: I'll squash all the commits before merging.

src/ServiceInsight/MessageHeaders/MessageHeadersView.xaml
HEskandari: This is moving the data templates to the upper scope (UserControl). The other issue is that the DesignTime data context doesn't support generics so the ReactiveList<T> is extracted into its own class and used. 

This allows the static analysis tools to find which class/property we have bound to.

src/ServiceInsight.Tests/packages.config
HEskandari: Yes in fact it is. We're not using it directly in Tests project. Admittedly this solves lots of Resharper warnings in the test project.

src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
HEskandari: There is another brush called SelectedBackgroundBrush which is also in the scope and was conflicting with this. They have different values though so I have changed the name of this one here.

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
distantcam: Could we agree to not put usings inside the namespace? I know the company in general does that but it causes all sorts of problems.

src/ServiceInsight.Tests/packages.config
distantcam: The other way to solve that is to turn off Resharper...

src/ServiceInsight/Shell/About.xaml
adamralph: 👍 thanks for the explanation.

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
adamralph: @HEskandari what is your preference? It seems that both @distantcam and I prefer full namespaces. Would you be OK with that?

Personally I prefer `using` inside `namespace` so I'd prefer to leave that as-is.

src/ServiceInsight.Tests/packages.config
adamralph: If not having the dependency listed here is problematic then I'm not strongly opposed to it.

@HEskandari before doing that, would you mind investigating if you can solve your local tooling (Resharper) problem in another way? E.g. by switching off that warning by default?

src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
adamralph: 👍 makes sense

src/ServiceInsight/MessageHeaders/MessageHeadersView.xaml
adamralph: Thanks for the explanation, the generic issue make sense and I understand that fix now.

> This is moving the data templates to the upper scope (UserControl).

What is the reason for moving them to the upper scope?

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
adamralph: I'm not sure that squashing all these commits into one is best. The refactorings each touch very disconnected parts of the code base so I'd prefer to keep them separate.

This can be fixed up by adding two fixup! commits for each of the offending commits.

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
distantcam: @adamralph usings inside the namespace is why we need to use the global:: root namespace. From a technical perspective, it's designed to protect the codebase from bad choices (like naming classes with the same names as base classes), but I'd rather have the code break if we do something dumb.

HEskandari: @adamralph @distantcam I'm okay with full namespaces. Should we ensure all the namespaces are full in this PR? or do it separately? is there a way to enforce it?

distantcam: @HEskandari I'd do it in a separate PR, along with a code rule to enforce it. Don't worry about it for this PR.

src/ServiceInsight.Tests/packages.config
HEskandari: @adamralph I tried updating the Resharper to see if it goes away but it didn't. These warnings also would show up in TeamCity if we turn on the inspections (which not surprisingly is disabled in SI build).

I think the only way to utterly mute it would be to add the exclusion of this specific warning to .DotSetting file.

HEskandari:
@adamralph I have made some changes so that each commit successfully compiles now.

HEskandari:
@Particular/serviceinsight-maintainers anything else needed to do here? Have I missed anything?

adamralph:
@HEskandari I'll review again ASAP.

adamralph:
@HEskandari I'd like to go through this PR with you. Please ping me when you're back online.

adamralph:
Discussed with @HEskandari - the "Clean up usings" commit needs to be trimmed down and then this should be good to merge.

HEskandari:
@adamralph The commit is now updated. I've removed all the unrelated code as well. Please review.

src/ServiceInsight.Tests/MessageBodyViewModelTests.cs
```diff
@@ -31,15 +30,14 @@ public class MessageBodyViewModelTests
         [SetUp]
         public void TestInitialize()
         {
-            eventAggregator = Substitute.For<IEventAggregator>();
             workNotifier = Substitute.For<IWorkNotifier>();
             serviceControl = Substitute.For<IServiceControl>();
             hexContent = Substitute.For<HexContentViewModel>();
             jsonContent = Substitute.For<JsonMessageViewModel>();
             xmlContent = Substitute.For<XmlMessageViewModel>();
             selection = new MessageSelectionContext(eventAggregator);
```
adamralph: This line doesn't compile anymore, since `eventAggregator` has been removed.

PR #667, opened 2016-12-06
Removing unused code from SagaView

distantcam:
As part of #657 I discovered 2 icons that did not work properly. This lead me to investigating the SagaView and I found there was a bit of code that was related to showing endpoints, except the bool controlling this code is never changed, and so all this xaml is never used.


distantcam:
This PR is being reverted until further testing can be done.

PR #674, opened 2016-12-08
Updates the saga view styles

### LINKED ISSUE #666: Improvements to property display styles in Saga View
There are a few issues with the way property key / values are displayed in SagaView so the following improvements should be made:

image

image

- Contrast for property names and values is to be increased while keeping special characters de-emphasized. 
- Show the active tab in underlined black text to indicate it's active and can't be clicked again.
- Mouse cursor should change to Hand when hovering over the links.
- In order to not have View properties: displaying as if it was another property and better align the display of custom tabs with SP, change tabs labels, reduce font size by 2pt and make it all caps.

HEskandari:
Resolves #666

- Added mouse cursor
- Changed the checked-tabs style to have underline (border) for both selected and mouse-hover states.
- Updated the colors/brushes as per design guide

The only thing left is the color of underline links when not selected.

### DIFF
--- a/src/ServiceInsight/Saga/SagaUpdateControl.xaml
+++ b/src/ServiceInsight/Saga/SagaUpdateControl.xaml
@@ -5,7 +5,6 @@
              xmlns:local="clr-namespace:ServiceInsight.Saga"
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:models="clr-namespace:ServiceInsight.Models"
-             xmlns:system="clr-namespace:System;assembly=mscorlib"
              Background="White"
              d:DataContext="{d:DesignInstance {x:Type local:SagaUpdate}}"
              d:DesignHeight="900"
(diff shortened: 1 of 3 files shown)


HEskandari:

HEskandari:
@adamralph the radio button 'links' are complex because out of the box WPF automatically reverses the bindings which means you can not have the link underlined both on mouse-hover and when selected. To achieve that effect simplest solution was to go with two borders and show/hide them. An underlined textblock although is simpler, it would not work exactly the way we wanted here.

I'll change the payload links as requested.

src/ServiceInsight/Saga/SagaUpdateControl.xaml
```diff
@@ -377,35 +401,28 @@
                                    Foreground="{StaticResource SagaForeground}"
                                    Text="{Binding Name}"
                                    TextTrimming="CharacterEllipsis" />
-                        <TextBlock Foreground="{StaticResource SagaForeground}" Text=" =    " />
-                        <TextBlock FontWeight="Bold"
-                                   Foreground="{StaticResource SagaForeground}"
+                        <TextBlock Foreground="{StaticResource SagaSymbolsValue}" Text=" =    " />
+                        <TextBlock Foreground="{StaticResource SagaForeground}"
```
HEskandari: @adamralph good point. I think that style doesn't provide any more value since we've removed other stuff. I think it's better to go back to just setting the foreground. I'll remove the style and make sure the usage is consistent.

HEskandari:
@adamralph The problem with that approach is that Textblock is not "selectable". We still need a way of selecting a view. Only having a textblock would not work. We could however, retemplate a control like button, check button, checkbox, or like this case a radio button. That would give us the selection and we can make it look like a Textblock which is what we have done.

The complexity, in my opinion, is because of the two borders inside the Radiobutton template. We could style the textblock to be underlined on mouse hover. We could also make the textblock underlined when the Radiobutton is selected. What's not possible - which has led us down the path of having two borders - is having both hover underline and selected underline.

adamralph:
Discussed this with @HEskandari, who will investigate an alternate approach to the radio buttons.

HEskandari:
Implemented a simpler approach suggested by @adamralph which is showing/hiding textblocks in the radio button template. The solution is *a lot* simpler now. Please review the changes made in commit `4b83ce5` @Particular/serviceinsight-maintainers

src/ServiceInsight/Saga/SagaUpdateControl.xaml
```diff
@@ -23,108 +22,38 @@
...
-                        <LinearGradientBrush StartPoint="0.5,0" EndPoint="0.5,1">
-                            <GradientStop Offset="0" Color="#FFA3AEB9" />
-                            <GradientStop Offset="0.375" Color="#FF8399A9" />
-                            <GradientStop Offset="0.375" Color="#FF718597" />
-                            <GradientStop Offset="1" Color="#FF617584" />
-                        </LinearGradientBrush>
-                    </Setter.Value>
-                </Setter>
+                <Setter Property="Padding" Value="0,0,0,1" />
+                <Setter Property="FontSize" Value="10" />
```
adamralph: Should this explicitly say `10px`? There's some inconsistency in the file, but in most instances `px` is stated explicitly.

HEskandari: As discussed the default value is in px. We decided to leave it as is for now.

Issue #639, opened 2016-11-21
Search strings with a forward slash followed by a numeric do not work


ramonsmits:
*Depends on https://github.com/Particular/ServiceControl/issues/921*

Feature request via support:

> In ServiceInsight there is now way to search messages when searched value contains slash. In our systems formats of various kinds identifiers have been created using "/" and the inability to find them in the body of the message is a big problem.
>
> I found a workaround using the REST API:
>
> http://localhost:33333/api/messages/search?q="T/432260"

adamralph:
@ramonsmits thanks for raising this. I guess we're just not escaping the string.

Out of interest, can you please check with the user to see if this works?

http://localhost:33333/api/messages/search?q=T%2F432260

ramonsmits:
@adamralph You mean entering "T%2F432260" in the ServiceInsight search box?

adamralph:
@ramonsmits no, I mean just hitting that URL.

If this is indeed a bug caused by lack of escaping, then entering the escaped string into the textbox might be a workaround until we fix the bug, but after we fix the bug, it will no longer work.

ramonsmits:
@Particular/serviceinsight-maintainers Has there been any progress on this?

adamralph:
@ramonsmits what was the result of trying to request http://localhost:33333/api/messages/search?q=T%2F432260 ?

ramonsmits:
Customer response:

> No, this solution does not work. I use RESTClient plugin for Firefox instead of ServiceInsight to search messages based on their id.
> Sample queries that’s works:
> - http://localhost:33333/api/messages/search?q="T/992685"
> - http://localhost:33333/api/messages/search?q=T/992685

adamralph:
@ramonsmits thanks, I'll investigate further.

adamralph:
It turns out that the problem is much more specific. See the updated version of https://github.com/Particular/ServiceControl/issues/921

@Particular/serviceinsight-maintainers in light of this new information, we can fix this immediately in SI, for all versions of SC, by switching to the query string API. I've raised #687 which does this. Strictly speaking, this is not a bug in SI, but rather SC, although since the SC API is for internal use only, I think we can regard this as a bug in SI in that it is making use of a broken internal API when a working internal API (the query string variant) exists. I propose we mark this issue a bug. Any objections?

PR #686, opened 2017-02-21
Fix backwards compatibility with ServiceControl

distantcam:
Related to #639


distantcam:
@adamralph And as discussed, I disagree, we can and should fix the bug now, rather than waiting.

PR #689, opened 2017-02-24
Changes to saga text styles and null conversion

### LINKED ISSUE #642: Null values are not represented in SagaView

HEskandari:
Fixes #642

### DIFF
--- a/src/ServiceInsight/Saga/SagaResources.xaml
+++ b/src/ServiceInsight/Saga/SagaResources.xaml
@@ -25,4 +25,13 @@
     
     <SolidColorBrush x:Key="FailedMessageBackground" Color="#F9916F" />
     <SolidColorBrush x:Key="RetriedMessageBackground" Color="#F2F2F2" />
+
+    <Style x:Key="ItalicNullTextStyle" TargetType="{x:Type TextBlock}">
+        <Setter Property="Foreground" Value="{StaticResource SagaForeground}" />
+        <Style.Triggers>
+            <Trigger Property="Tag" Value="{x:Null}">
+                <Setter Property="FontStyle" Value="Italic" />
+            </Trigger>
+        </Style.Triggers>
+    </Style>
 </ResourceDictionary>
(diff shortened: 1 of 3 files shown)


HEskandari:

src/ServiceInsight/Saga/SagaResources.xaml
```diff
@@ -25,4 +25,13 @@
     
     <SolidColorBrush x:Key="FailedMessageBackground" Color="#F9916F" />
     <SolidColorBrush x:Key="RetriedMessageBackground" Color="#F2F2F2" />
+
+    <Style x:Key="GrayNullTextStyle" TargetType="{x:Type TextBlock}">
```
distantcam: This style doesn't make the text gray when it's null. Should the name be changed? Something more generic like `ValueTextStyle`

HEskandari: Yes, good point. It is more like 'ItalicWhenNull' style due to the Trigger. I'll rename it.

HEskandari: @distantcam This is done with the latest push. Please review.

distantcam: I'm not a fan of the new name either. What if the style changes to bold?

I prefer a functional name, like `SagaValueTextStyle`. Thoughts @adamralph?

HEskandari: If it changes to bold, should we question it then and rename accordingly? I'm open to other names but this is not a generic text style so don't think `SagaValueTextStyle` would be a good choice.

adamralph: I prefer a functional name too. I guess we can always refactor later if we feel strongly enough.

PR #690, opened 2017-02-27
Search using RestSharp parameters

### LINKED ISSUE #639: Search strings with a forward slash followed by a numeric do not work

distantcam:
A followup to #687 to use RestSharp's parameters instead of string manipulation.

Fixes #639.

### DIFF
--- a/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
+++ b/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
@@ -13,7 +13,6 @@
     using Caliburn.Micro;
     using Framework;
     using RestSharp;
-    using RestSharp.Contrib;
     using RestSharp.Deserializers;
     using Serilog;
     using ServiceInsight.ExtensionMethods;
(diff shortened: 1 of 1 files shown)


src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -210,7 +209,8 @@ void AppendSearchQuery(IRestRequest request, string searchQuery)
                 return;
             }
 
-            request.Resource += string.Format("search?q={0}", HttpUtility.UrlEncode(searchQuery));
+            request.Resource += "search";
+            request.AddParameter("q", searchQuery, ParameterType.GetOrPost);
```
HEskandari: You also removed `HttpUtility.UrlEncode` from this line. Does AddParameter does the encoding internally?

distantcam: Yes it handles encoding the name (in our case 'q') as well as the parameter.

adamralph:
Thanks @distantcam.

To make things simpler, I removed the label and milestone and connected the PR to #639 as the "proper" fix for that issue.

PR #691, opened 2017-03-01
[WIP] Disabled shared compilation

distantcam:


src/ServiceInsight/ServiceInsight.csproj
```diff
@@ -45,9 +46,6 @@
     <UseVSHostingProcess>false</UseVSHostingProcess>
     <CodeAnalysisRuleSet>..\CommonRuleSet.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
-  <PropertyGroup>
-    <ApplicationIcon>ServiceInsight.ico</ApplicationIcon>
```
HEskandari: Was this intentionally removed?

src/ServiceInsight/ServiceInsight.csproj
```diff
@@ -21,6 +21,7 @@
     <NuGetPackageImportStamp>
     </NuGetPackageImportStamp>
     <ExpressionBlendVersion>12.0.50429.0</ExpressionBlendVersion>
+    <UseSharedCompilation>false</UseSharedCompilation>
```
HEskandari: Is this disabled only for Debug config? How about Release?

src/ServiceInsight/ServiceInsight.csproj
distantcam: It's already set in a previous property group. This is redundant.

src/ServiceInsight/ServiceInsight.csproj
distantcam: It's disabled for both.

distantcam:
@Particular/serviceinsight-maintainers This was to test if this would fix some errors occurring in the build. Instead the build was fixed with a parameter so this PR is no longer needed. Closing.

PR #692, opened 2017-03-01
Using the WindowManager to show a dialog for SagaContentViewer

### LINKED ISSUE #643: 'Detail view' popup is not behaving as an standard window or popup.
- Replacing the detail view popup with a standard resizable window that behaves and looks the same as other dialogs (e.g. connection dialog). 
- The window should not remain as the top window when switching to other Windows applications.
- The window should be non-modal (but still sit above the main app window).

This is a split of #624.

### LINKED ISSUE #644: Scrollbars are behaving inconsistently in 'Detail view'
- Unnecessary scrollbars should be hidden. 

This is a split of #624.

HEskandari:
Fixes #643 
Fixes #644


src/ServiceInsight/MessagePayloadViewer/MessagePayloadView.xaml
```diff
@@ -0,0 +1,40 @@
...
+             xmlns:controls="clr-namespace:ServiceInsight.Controls"
+             Width="800"
+             Height="400"
+             Background="#FF252526"
+             mc:Ignorable="d" 
+             d:DesignHeight="500"
+             d:DesignWidth="900"
+             d:DataContext="{d:DesignInstance {x:Type local:MessagePayloadView}}">
+    <UserControl.Resources>
+    </UserControl.Resources>
```
distantcam: This can be cleaned up.

src/ServiceInsight/MessagePayloadViewer/MessagePayloadViewModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace ServiceInsight.MessagePayloadViewer
+{
+    using Caliburn.Micro;
+
+    public class MessagePayloadViewModel : Screen
+    {
+        public string Content { get; private set; }
```
distantcam: With C# 6 the `private set` can be removed.

HEskandari:
@distantcam thanks for your comments. They're all addressed. Please have another look and see if there's anything else.

HEskandari:
@distantcam Fixed the window minimization, that shouldn't happen anymore. With that the resizing is a non-issue.

distantcam:
@adamralph Did you want to review this?

adamralph:
@distantcam I did want to yes, but first, time for a stupid question: how do I get this "Detail view" to show?

adamralph:
BTW - is the title of the PR accurate? The only mentions I see of `SagaContentViewer` are removals, and it looks like the resulting code doesn't have any such object in it.

HEskandari:
@adamralph there's no stupid questions :) this works as a replacement of the old dialog. You can get to it by clicking a 'link' on the saga view, e.g.:

image

distantcam:
@adamralph I had the same issue, in that I couldn't get the right data into SI. 

If you run the SagaAudit acceptence tests https://github.com/Particular/ServiceControl.Plugin.Nsb6.SagaAudit/tree/master/src/ServiceControl.Plugin.Nsb6.SagaAudit.AcceptanceTests it will add a bunch of messages with the data you need.

If you have trouble building it like I did, turn off signing of the assemblies.

HEskandari:
@distantcam @adamralph Just a side note that you need a proper message with a large payload to see how the wrapping works. The sample message payloads generated by the smoke test project are not big enough AFAIR.

adamralph:
OK, so I managed to smoke test it by hacking a great big object onto a property of the saga data in the audit plugin smoke test.

Wow, the new window is so much better.

One thing to note though is that the new window text-wraps the property values. Is that OK? I'm guessing it is OK, and it's probably desirable, but I just wanted to check.

### Old

image

### New

image

adamralph:
BTW - is it easy to make the window close when <kbd>Esc</kbd> is pressed?

HEskandari:
@adamralph The wrapping is intentional, as it is the only way to fix the weird scrollbar issues mentioned in #644.

RE the <kbd>Esc</kbd> key, should that behave like that for all the modal dialogs? (I'd say yes for consistency reasons).

distantcam:
Should making <kbd>Esc</kbd> work with all dialogs be a separate PR? In which case this one should be merged.

adamralph:
Discussed with @HEskandari. Spun off closing of all dialogs with <kbd>Esc</kbd> into https://github.com/Particular/ServiceInsight/issues/711

sergioc:
> image

I think this view could use a bit of padding around the window borders - otherwise the content looks too tight crowded.

As for text wrapping, code editors don't wrap text AFAIK. Won't the wrapping give the illusion there's multiple lines of text when that is not true? How about making wrapping an option instead?

adamralph:
Let's re-open #643 and move the conversation there.

PR #694, opened 2017-03-06
update the release guide after observations made during 1.6.2 release

adamralph:


adamralph:
For reference, we discussed the 1.6.2 milestone in the SI sync today and we're not sure why it wasn't closed as part of the 1.6.2 release. The new section in the document contains a check to see if the milestone is closed.

distantcam:
The milestone not closing is due to the deployment server switching from our GitReleaseManager to the GitTools one. I'll go fix that separately, but that shouldn't affect this PR.

👍

Issue #664, opened 2016-12-06
InvalidOperationException Sequence contains no matching element in 1.5.10


distantcam:
This was reported via Raygun

ServiceInsight 1.5.10
ServiceControl 1.11.0

    System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate):77
     ServiceInsight.ServiceControl.DefaultServiceControl+<>c__31`1.<GetPagedResult>b__31_0(IRestResponse`1 response) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\ServiceControl\DefaultServiceControl.cs:245
     ServiceInsight.ServiceControl.DefaultServiceControl.ProcessResponse[T,T2](Func`2 selector, IRestResponse`1 response) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\ServiceControl\DefaultServiceControl.cs:459
     ServiceInsight.ServiceControl.DefaultServiceControl.Execute[T,T2](RestRequestWithCache request, Func`2 selector) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\ServiceControl\DefaultServiceControl.cs:398
     ServiceInsight.ServiceControl.DefaultServiceControl.Search(String searchQuery, Int32 pageIndex, String orderBy, Boolean ascending) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\ServiceControl\DefaultServiceControl.cs:99
     ServiceInsight.MessageList.MessageListViewModel.RefreshMessages(Endpoint endpoint, Int32 pageIndex, String searchQuery, String orderBy, Boolean ascending) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\MessageList\MessageListViewModel.cs:148
     ServiceInsight.MessageList.MessageListViewModel.RefreshMessages(String orderBy, Boolean ascending) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\MessageList\MessageListViewModel.cs:109
     ServiceInsight.MessageList.MessageListViewModel.OnSelectedExplorerItemChanged() in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\MessageList\MessageListViewModel.cs:235
[...]

PR #696, opened 2017-03-07
Handling null values for total-count header.

distantcam:
Fix for #664

### DIFF
--- a/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
+++ b/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
@@ -236,23 +236,25 @@
             return client;
         }
 
-        static RestRequestWithCache CreateMessagesRequest(string endpointName = null) => endpointName != null
-    ? new RestRequestWithCache(string.Format(EndpointMessagesEndpoint, endpointName), RestRequestWithCache.CacheStyle.IfNotModified)
-    : new RestRequestWithCache(MessagesEndpoint, RestRequestWithCache.CacheStyle.IfNotModified);
+        static RestRequestWithCache CreateMessagesRequest(string endpointName = null) =>
+            endpointName != null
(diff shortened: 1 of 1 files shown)


distantcam:
@Particular/serviceinsight-maintainers So how do we feel about this one?

HEskandari:
I'd say if we have no idea how this happens and the header is always expected to be there and has been there in every release of SC, there's no reason to write defensive code around it.

If anything, we need to throw a more meaningful exception highlighting the problem, because with no total-count, other stuff like paging are going to get broken as well.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -236,23 +236,25 @@ IRestClient CreateClient(string baseUrl = null)
...
+            : new RestRequestWithCache(MessagesEndpoint, RestRequestWithCache.CacheStyle.IfNotModified);
 
         PagedResult<T> GetPagedResult<T>(RestRequestWithCache request) where T : class, new()
         {
             var result = Execute<PagedResult<T>, List<T>>(request, response => new PagedResult<T>
             {
                 Result = response.Data,
-                TotalCount = int.Parse(response.Headers.First(x => x.Name == ServiceControlHeaders.TotalCount).Value.ToString())
+                TotalCount = int.Parse(
+                    response.Headers.FirstOrDefault(x => x.Name == ServiceControlHeaders.TotalCount)?.Value?.ToString() ?? "0")
```
adamralph: Is this line the only relevant change?

I.e. is the rest of this commit just whitespace fixes?

adamralph:
I tend to agree that we should throw a specific exception when we are not told the total count.

distantcam:
Silently trying to recover from this failure isn't the right solution. Closing this PR.

PR #698, opened 2017-03-13
Cleanup of some manual notify calls

distantcam:
While reviewing adding more reactive stuff I noticed a few bad `NotifyOfPropertyChange` calls.

### DIFF
--- a/src/ServiceInsight/MessageList/MessageListViewModel.cs
+++ b/src/ServiceInsight/MessageList/MessageListViewModel.cs
@@ -37,6 +37,7 @@
         string lastSortColumn;
         bool lastSortOrderAscending;
         IMessageListView view;
+        ExplorerItem selectedExplorerItem;
 
         public MessageListViewModel(
             IEventAggregator eventAggregator,
@@ -79,8 +80,6 @@
 
         public bool WorkInProgress => WorkCount > 0 && !Parent.AutoRefresh;
 
-        public ExplorerItem SelectedExplorerItem { get; private set; }
-
         public ICommand RetryMessageCommand { get; }
 
         public ICommand CopyMessageIdCommand { get; }
@@ -100,7 +99,7 @@
 
         public void RefreshMessages(string orderBy = null, bool ascending = false)
         {
-            var serviceControlExplorerItem = SelectedExplorerItem as ServiceControlExplorerItem;
+            var serviceControlExplorerItem = selectedExplorerItem as ServiceControlExplorerItem;
             if (serviceControlExplorerItem != null)
             {
                 RefreshMessages(searchQuery: SearchBar.SearchQuery,
@@ -109,7 +108,7 @@
                                      ascending: ascending);
             }
 
-            var endpointNode = SelectedExplorerItem as AuditEndpointExplorerItem;
+            var endpointNode = selectedExplorerItem as AuditEndpointExplorerItem;
             if (endpointNode != null)
             {
                 RefreshMessages(searchQuery: SearchBar.SearchQuery,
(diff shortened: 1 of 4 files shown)


src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -76,7 +75,9 @@ public class MessageListViewModel : RxConductor<RxScreen>.Collection.AllActive,
 
         public MessageSelectionContext Selection { get; }
 
-        public bool WorkInProgress => workCount > 0 && !Parent.AutoRefresh;
+        public int WorkCount { get; private set; }
```
HEskandari: The previous implementation had this `WorkCount` as internal. While it is made public is it used anywhere? AFAIK all we bind to is `WorklInProgress` propery.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -36,7 +36,6 @@ public class MessageListViewModel : RxConductor<RxScreen>.Collection.AllActive,
         GeneralHeaderViewModel generalHeaderDisplay;
         string lastSortColumn;
         bool lastSortOrderAscending;
-        int workCount;
```
adamralph: This change seems redundant, and unrelated to introducing `nameof`

src/ServiceInsight/MessageList/MessageListViewModel.cs
distantcam: Yep. I thought I had to make it public for the Fody magic to work, but it turns out it works on private properties too. I'll modify the PR.

src/ServiceInsight/MessageList/MessageListViewModel.cs
distantcam: This PR isn't just about introducing nameof. It's also about removing unnecessary `NotifyOfPropertyChange` calls. I plan on eliminating as many as I can, but that's enough for this PR.

As for this change, it's not redundant. By making it a property `PropertyChanged.Fody` weaves in the change notification automatically. `PropertyChanged.Fody` also detects any properties using `WorkCount` like `WorkInProgress` and automatically fires a notification for that property too. That's how I can eliminate the manual `NotifyOfPropertyChange(WorkInProgress)` calls.

adamralph: Ah, IL weaving magic 😄 

Thanks for the explanation.

adamralph:
@HEskandari do you have any remaining concerns?

HEskandari:
@adamralph yes :) I initially thought the private field is better, but had a chat with @distantcam and we agreed that a public property is a better choice here. The reasoning is that for someone new to the code it is hard to understand why a private property is needed. Also the tooling (e.g. Resharper and alikes) would wrongly suggest you to remove it.

distantcam:
@HEskandari You meant private property, not private field. @adamralph The discussion Hadi and I had was on private vs public, not on field vs property.

HEskandari:
@distantcam yes, correct. Sorry for the confusion.

We decided to change it back to public property AFAIR and not a field.

distantcam:
@HEskandari Yep. I've rolled back the last commit, and rebrached everything onto latest master.

HEskandari:
@distantcam cool, I'm seeing it now.

adamralph:
Discussed with @HEskandari. He will smoke test this and if all OK, merge it.

HEskandari:
@adamralph I just got to this today. I've smoke tested and things are working as expected.

PR #705, opened 2017-03-17
Clearing saga view and message body view when the selected endpoint h…

### LINKED ISSUE #625: When selecting the endpoint in Endpoint Explorer the loaded body doesn't get cleared
Steps to reproduce:

- Connect to a ServiceControl where you have endpoints with and without message
- Switch to the endpoint which has messages.
- Select any message, move to the body tab. See the message body.
- On Endpoint Explorer switch to the other endpoint that has no messages.

Observe that the old message body will still appear, which is incorrect. Also, when you select this endpoint, all three body tabs (Xml, Json, Hex) appear regardless of the message content type.

### LINKED ISSUE #700: Switching to an endpoint with no message some views are not cleared
Some views are working as expected but others are not cleared when changing to an endpoint with no message. The following views needs to be fixed:

- Saga View
- Body tab views (Xml, Json, Hex)

To create an endpoint with no message, create an NSB endpoint with no sending/receiving messages and enable heart-beat plugin.

HEskandari:
Resolves #625

### DIFF
--- a/src/ServiceInsight/MessageViewers/MessageBodyViewModel.cs
+++ b/src/ServiceInsight/MessageViewers/MessageBodyViewModel.cs
@@ -87,6 +87,7 @@
 
         public void Handle(SelectedMessageChanged @event)
         {
+            ClearMessageDisplays();
             LoadMessageBody();
 
             if (selection.SelectedMessage != null)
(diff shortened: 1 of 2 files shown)


adamralph:
@HEskandari would it be correct to say that this fixes two bugs:

- #625 
- a similar bug for the saga view

HEskandari:
@adamralph there was an issue I had opened which was a duplicate of #625 but I have closed it since. Do you mean there's another similar bug other than that? Which one exactly?

adamralph:
As discussed, #625 has been altered to include saga view as well.

PR #706, opened 2017-03-17
Making the null values gray

### LINKED ISSUE #697: Null values in saga view are not grey
Related to #642

HEskandari:
Fixes #697


HEskandari:
@distantcam How about `PropertyValueTextStyle` as it is used to display the property values (and not links).

distantcam:
@HEskandari Yep, I'm fine with that too.

HEskandari:
@distantcam Thanks. Updated.

PR #737, opened 2017-07-11
Use SC-generated paging links to page the audit message data

### LINKED ISSUE #736: Connecting to ServiceControl does not trigger a refresh of the message list when the endpoint explorer does not have focus
<details>
<summary>Original description</summary>
If I don't select anything in the `Endpoint Explorer` changing the sort order and/or column does not trigger query to SC to fetch the data. The sorting is done on the local data set. Only when I first click something in the `Endpoint Explorer` the sorting starts to work properly.

I suspect this is because of this code https://github.com/Particular/ServiceInsight/blob/master/src/ServiceInsight/MessageList/MessageListViewModel.cs#L102
</details>

SzymonPobiega:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1650

SC generates `Link` header which contains up to 4 links: next, prev, first and last. These are used, instead of generating the links in SI, to page through the data.

The main benefit is making SI not dependent on the actual paging URL format of SC. The SI only needs to know how to initiate the search.

The actual goal is to be able to insert an intermediary between SI and SC. That intermediary could aggregate the results from multiple instances of SC without SI even knowing about its existence.

Also contains small refactorings to SC gateway.

Also solves #736

### Page Count

As part of the TF work of extending SC with the multi instance mode for audits we discovered that the page count is not possible to determine upfront by doing a simple calculation based on the assumed page size. Since we went from a simplified paging approach in contrast to keep the offsets per SC instance each page can return up to `numberOfInstance * page_size` worth of data. Audits cannot be assumed to be uniformly distributed. Based on that we figured having a page size visible at all time makes no sense. The ability to page through the data is more important than knowing how many pages there are.

---

Fixes #736

### DIFF
--- a/src/ServiceInsight.sln.DotSettings
+++ b/src/ServiceInsight.sln.DotSettings
@@ -562,6 +562,7 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateThisQualifierSettings/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002EJavaScript_002ECodeStyle_002ESettingsUpgrade_002EJsCodeFormatterSettingsUpgrader/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002EJavaScript_002ECodeStyle_002ESettingsUpgrade_002EJsParsFormattingSettingsUpgrader/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002EJavaScript_002ECodeStyle_002ESettingsUpgrade_002EJsWrapperSettingsUpgrader/@EntryIndexedValue">True</s:Boolean>
 	
 	
--- a/src/ServiceInsight/MessageList/MessageListView.xaml.cs
+++ b/src/ServiceInsight/MessageList/MessageListView.xaml.cs
@@ -60,7 +60,10 @@
 
         void SortData(ColumnBase column, ListSortDirection order)
         {
-            Model.RefreshMessages(column.Tag as string, order == ListSortDirection.Ascending);
+            var orderBy = column.Tag as string;
+            var ascending = order == ListSortDirection.Ascending;
+
+            Model.RefreshMessages(orderBy, ascending);
         }
 
         void Grid_OnCustomColumnDisplayText(object sender, CustomColumnDisplayTextEventArgs e)
--- a/src/ServiceInsight/MessageList/MessageListViewModel.cs
+++ b/src/ServiceInsight/MessageList/MessageListViewModel.cs
@@ -151,7 +151,6 @@
                     return;
                 }
 
-                pagedResult.CurrentPage = 1;
                 TryRebindMessageList(pagedResult);
 
                 SearchBar.IsVisible = true;
--- a/src/ServiceInsight/MessageList/MessageListViewModel.cs
+++ b/src/ServiceInsight/MessageList/MessageListViewModel.cs
@@ -37,8 +37,8 @@
         IWorkNotifier workNotifier;
         IServiceControl serviceControl;
         GeneralHeaderViewModel generalHeaderDisplay;
-        string lastSortColumn;
-        bool lastSortOrderAscending;
+        string sortColumn;
+        bool sortOrderAscending;
         IMessageListView view;
         ExplorerItem selectedExplorerItem;
 
@@ -104,76 +104,59 @@
             clipboard.CopyTo(generalHeaderDisplay.HeaderContent);
         }
 
-        public void RefreshMessages(string orderBy = null, bool ascending = false)
+        public void NavigateToPage(string link, int pageIndex)
         {
-            var serviceControlExplorerItem = selectedExplorerItem as ServiceControlExplorerItem;
-            if (serviceControlExplorerItem != null)
+            using (workNotifier.NotifyOfWork("Loading messages..."))
             {
-                RefreshMessages(searchQuery: SearchBar.SearchQuery,
-                                     endpoint: null,
-                                     orderBy: orderBy,
-                                     ascending: ascending);
-            }
-
-            var endpointNode = selectedExplorerItem as AuditEndpointExplorerItem;
-            if (endpointNode != null)
-            {
-                RefreshMessages(searchQuery: SearchBar.SearchQuery,
-                                     endpoint: endpointNode.Endpoint,
-                                     orderBy: orderBy,
-                                     ascending: ascending);
-            }
-        }
-
-        public void RefreshMessages(Endpoint endpoint, int pageIndex = 1, string searchQuery = null, string orderBy = null, bool ascending = false)
-        {
-            using (workNotifier.NotifyOfWork($"Loading {(endpoint == null ? "all" : endpoint.Address)} messages..."))
-            {
-                if (orderBy != null)
-                {
-                    lastSortColumn = orderBy;
-                    lastSortOrderAscending = ascending;
-                }
-
-                PagedResult<StoredMessage> pagedResult;
-
-                if (endpoint != null)
-                {
-                    pagedResult = serviceControl.GetAuditMessages(endpoint,
-                        pageIndex: pageIndex,
-                        searchQuery: searchQuery,
-                        orderBy: lastSortColumn,
-                        ascending: lastSortOrderAscending);
-                }
-                else if (!searchQuery.IsEmpty())
-                {
-                    pagedResult = serviceControl.Search(pageIndex: pageIndex,
-                        searchQuery: searchQuery,
-                        orderBy: lastSortColumn,
-                        ascending: lastSortOrderAscending);
-                }
-                else
-                {
-                    pagedResult = serviceControl.Search(pageIndex: pageIndex,
-                        searchQuery: null,
-                        orderBy: lastSortColumn,
-                        ascending: lastSortOrderAscending);
-                }
+                var pagedResult = serviceControl.GetAuditMessages(link);
 
                 if (pagedResult == null)
                 {
                     return;
                 }
 
+                pagedResult.CurrentPage = pageIndex;
                 TryRebindMessageList(pagedResult);
 
                 SearchBar.IsVisible = true;
-                SearchBar.SetupPaging(new PagedResult<StoredMessage>
+                SearchBar.SetupPaging(pagedResult);
+            }
+        }
+
+        public void RefreshMessages(string orderBy, bool ascending)
+        {
+            sortColumn = orderBy;
+            sortOrderAscending = ascending;
+            RefreshMessages();
+        }
+
+        public void RefreshMessages()
+        {
+            var endpointNode = selectedExplorerItem as AuditEndpointExplorerItem;
+
+            Search(searchQuery: SearchBar.SearchQuery, endpoint: endpointNode?.Endpoint);
+        }
+
+        public void Search(Endpoint endpoint, string searchQuery = null)
+        {
+            using (workNotifier.NotifyOfWork($"Loading {(endpoint == null ? "all" : endpoint.Address)} messages..."))
+            {
+                var pagedResult = serviceControl.GetAuditMessages(
+                    endpoint,
+                    searchQuery,
+                    sortColumn,
+                    sortOrderAscending);
+
+                if (pagedResult == null)
                 {
-                    CurrentPage = pagedResult.CurrentPage,
-                    TotalCount = pagedResult.TotalCount,
-                    Result = pagedResult.Result,
-                });
+                    return;
+                }
+
+                pagedResult.CurrentPage = 1;
+                TryRebindMessageList(pagedResult);
+
+                SearchBar.IsVisible = true;
+                SearchBar.SetupPaging(pagedResult);
             }
         }
 
--- a/src/ServiceInsight/MessageList/MessageListViewModel.cs
+++ b/src/ServiceInsight/MessageList/MessageListViewModel.cs
@@ -104,7 +104,7 @@
             clipboard.CopyTo(generalHeaderDisplay.HeaderContent);
         }
 
-        public void NavigateToPage(string link, int pageIndex)
+        public void NavigateToPage(string link)
         {
             using (workNotifier.NotifyOfWork("Loading messages..."))
             {
@@ -115,7 +115,6 @@
                     return;
                 }
 
-                pagedResult.CurrentPage = pageIndex;
                 TryRebindMessageList(pagedResult);
 
                 SearchBar.IsVisible = true;
--- a/src/ServiceInsight/Search/SearchBarView.xaml
+++ b/src/ServiceInsight/Search/SearchBarView.xaml
@@ -69,7 +69,6 @@
             <Button Content="3" cal:Message.Attach="[Event Click]=[Action GoToPreviousPage]" />
             <TextBlock Text="Page" />
             <TextBlock Text="{Binding CurrentPage}" />
-            <TextBlock Text="{Binding PageCount, StringFormat=of {0}}" />
             <Button Content="4" cal:Message.Attach="[Event Click]=[Action GoToNextPage]" />
             <Button Content=":" cal:Message.Attach="[Event Click]=[Action GoToLastPage]" />
         </StackPanel>
--- a/src/ServiceInsight/Search/SearchBarViewModel.cs
+++ b/src/ServiceInsight/Search/SearchBarViewModel.cs
@@ -51,22 +51,22 @@
 
         public void GoToFirstPage()
         {
-            Parent.NavigateToPage(FirstLink, 1);
+            Parent.NavigateToPage(FirstLink);
         }
 
         public void GoToPreviousPage()
         {
-            Parent.NavigateToPage(PrevLink, CurrentPage - 1);
+            Parent.NavigateToPage(PrevLink);
         }
 
         public void GoToNextPage()
         {
-            Parent.NavigateToPage(NextLink, CurrentPage + 1);
+            Parent.NavigateToPage(NextLink);
         }
 
         public void GoToLastPage()
         {
-            Parent.NavigateToPage(LastLink, PageCount);
+            Parent.NavigateToPage(LastLink);
         }
 
         public ICommand SearchCommand { get; }
@@ -135,19 +135,6 @@
 
         public new MessageListViewModel Parent => base.Parent as MessageListViewModel;
 
-        public int PageCount
-        {
-            get
-            {
-                if (TotalItemCount == 0 || PageSize == 0)
-                {
-                    return 0;
-                }
-
-                return (int)Math.Ceiling((double)TotalItemCount / PageSize);
-            }
-        }
-
         public bool WorkInProgress => workCount > 0;
 
         public Endpoint SelectedEndpoint { get; private set; }
@@ -198,7 +185,6 @@
 
         public void NotifyPropertiesChanged()
         {
-            NotifyOfPropertyChange(nameof(PageCount));
             NotifyOfPropertyChange(nameof(CanGoToFirstPage));
             NotifyOfPropertyChange(nameof(CanGoToLastPage));
             NotifyOfPropertyChange(nameof(CanGoToNextPage));
--- a/src/ServiceInsight/Search/SearchBarViewModel.cs
+++ b/src/ServiceInsight/Search/SearchBarViewModel.cs
@@ -32,7 +32,6 @@
         {
             this.commandLineArgParser = commandLineArgParser;
             this.settingProvider = settingProvider;
-            PageSize = 50; //NOTE: Do we need to change this?
 
             SearchCommand = Command.Create(this, Search, vm => vm.CanSearch);
             CancelSearchCommand = Command.Create(this, CancelSearch, vm => vm.CanCancelSearch);
@@ -52,22 +51,22 @@
 
         public void GoToFirstPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, 1, SearchQuery);
+            Parent.NavigateToPage(FirstLink, 1);
         }
 
         public void GoToPreviousPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, CurrentPage - 1, SearchQuery);
+            Parent.NavigateToPage(PrevLink, CurrentPage - 1);
         }
 
         public void GoToNextPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, CurrentPage + 1, SearchQuery);
+            Parent.NavigateToPage(NextLink, CurrentPage + 1);
         }
 
         public void GoToLastPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, PageCount, SearchQuery);
+            Parent.NavigateToPage(LastLink, PageCount);
         }
 
         public ICommand SearchCommand { get; }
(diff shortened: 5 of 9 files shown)


src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -537,6 +564,7 @@ void LogError(IRestResponse response)
         static bool HasSucceeded(IRestResponse response) => successCodes.Any(x => response != null && x == response.StatusCode && response.ErrorException == null);
 
         static IEnumerable<HttpStatusCode> successCodes = new[] { HttpStatusCode.OK, HttpStatusCode.Accepted, HttpStatusCode.NotModified, HttpStatusCode.NoContent };
+        static Regex linkExpression = new Regex(@"<([^>]+)>;\s?rel=""(\w+)"",?\s?", RegexOptions.Compiled);
```
HEskandari: Isn't string operations simple enough for this purpose? Whenever I see a regex I cringe.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -97,28 +97,49 @@ public void CopyHeaders()
...
+                if (pagedResult == null)
+                {
+                    return;
+                }
+
+                pagedResult.CurrentPage = pageIndex;
+                TryRebindMessageList(pagedResult);
+
+                SearchBar.IsVisible = true;
+                SearchBar.SetupPaging(pagedResult.TotalCount > 0 ? pagedResult.CurrentPage : 0, pagedResult.TotalCount, pagedResult.PageSize, pagedResult.Result,
```
HEskandari: This is too verbose IMO. More on it down below.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -97,28 +97,49 @@ public void CopyHeaders()
...
                                      orderBy: orderBy,
                                      ascending: ascending);
             }
+            else
+            {
+                InitSearch(searchQuery: SearchBar.SearchQuery,
+                    endpoint: null,
+                    orderBy: orderBy,
+                    ascending: ascending);
+            }
```
HEskandari: the branching seem to be very similar and duplicated? could we get away with one? e.g.

endpoint = endpointNode ? null : endpointNode.Endpoint;
InitSearch(....)

so no branching.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -161,12 +165,12 @@ public void RefreshMessages(Endpoint endpoint, int pageIndex = 1, string searchQ
                 TryRebindMessageList(pagedResult);
 
                 SearchBar.IsVisible = true;
-                SearchBar.SetupPaging(new PagedResult<StoredMessage>
-                {
-                    CurrentPage = pagedResult.CurrentPage,
-                    TotalCount = pagedResult.TotalCount,
-                    Result = pagedResult.Result,
-                });
+                SearchBar.SetupPaging(pagedResult.TotalCount > 0 ? pagedResult.CurrentPage : 0, pagedResult.TotalCount, pagedResult.PageSize, pagedResult.Result,
```
HEskandari: Could we simplify the SetupPaging method? too many arguments.

src/ServiceInsight/Search/SearchBarViewModel.cs
```diff
@@ -52,22 +51,22 @@ protected override void OnActivate()
 
         public void GoToFirstPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, 1, SearchQuery);
+            Parent.RefreshMessages(FirstLink, 1);
```
HEskandari: Why do we still need the current page? Isn't the First/Prev/Last/Next link enough regardless of which page we're in? Same goes for usages below.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -88,38 +90,37 @@ public Uri CreateServiceInsightUri(StoredMessage message)
...
-            var request = CreateMessagesRequest(endpoint.Name);
-
-            AppendSystemMessages(request);
-            AppendSearchQuery(request, searchQuery);
-            AppendPaging(request, pageIndex);
-            AppendOrdering(request, orderBy, ascending);
-
-            var result = GetPagedResult<StoredMessage>(request);
-            if (result != null)
+            if (link.StartsWith("http"))
```
HEskandari: StartWith("http", StringComparison.OrdinalIgnoreCase) ?
Why do we need to check this? The code also looks very similar. Could we do away with branching?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -197,9 +198,9 @@ void AppendOrdering(IRestRequest request, string orderBy, bool ascending)
             request.AddParameter("direction", ascending ? "asc" : "desc", ParameterType.GetOrPost);
         }
 
-        void AppendPaging(IRestRequest request, int pageIndex)
+        void AppendPaging(IRestRequest request, int pageSize)
```
HEskandari: is there anywhere where we change the page size? If it is a constant already and can't be changed, we can also remove the page sizes.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: There are two things that depend on the result of this comparison. I can replace the `if` with `?` but I would end up with two `?` operators.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: It has a default value in SC (same as here, 50 messages). Do you say we should just rely on the SC default here?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: Unfortunately not because `,` is used as a delimiting character  but can also be present in the URL, e.g. `http://localhost:33333/api/messages?offsets=0,1,2,3; rel="next", http://localhost:33333/....`.

src/ServiceInsight/Search/SearchBarViewModel.cs
SzymonPobiega: We need to display the page index in the UI. The current SC unfortunately does not include the current page index in the response but will include in future. Then we'll able to drop these.

SzymonPobiega:
@HEskandari I've pushed fixes to the issues you pointed out and some refactorings aimed to reduce the complexity.

SzymonPobiega:
Guys, can I ask you for another round of reviews?

HEskandari:
I can review and test it but only mid-week next week.

HEskandari:
@SzymonPobiega I did test it today, as promised. It seems to be working fine. I'm happy to merge this in.

src/ServiceInsight/Search/SearchBarViewModel.cs
```diff
@@ -122,11 +126,16 @@ public void ClearPaging()
...
         {
-            Parent.RefreshMessages(SelectedEndpoint, CurrentPage, SearchQuery);
+            if (SelfLink != null)
+            {
+                Parent.NavigateToPage(SelfLink, CurrentPage);
+            }
+            else
+            {
+                Parent.Search(SelectedEndpoint, SearchQuery);
+            }
```
adamralph: How is this method invoked? I see no reference to it.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
adamralph: I'd rather not have to maintain (and possibly debug) this regex. How complicated would the equivalent imperative code be?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -88,38 +90,30 @@ public Uri CreateServiceInsightUri(StoredMessage message)
...
-            if (result != null)
-            {
-                result.CurrentPage = pageIndex;
-            }
-            return result;
+            return GetPagedResult<StoredMessage>(request);
         }
 
-        public PagedResult<StoredMessage> GetAuditMessages(Endpoint endpoint, string searchQuery = null, int pageIndex = 1, string orderBy = null, bool ascending = false)
+        public PagedResult<StoredMessage> GetAuditMessages(string link)
```
adamralph: What's the logic behind the two conditions in this method?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -88,38 +90,30 @@ public Uri CreateServiceInsightUri(StoredMessage message)
 
         public SagaData GetSagaById(Guid sagaId) => GetModel<SagaData>(new RestRequestWithCache(string.Format(SagaEndpoint, sagaId), RestRequestWithCache.CacheStyle.IfNotModified)) ?? new SagaData();
 
-        public PagedResult<StoredMessage> Search(string searchQuery, int pageIndex = 1, string orderBy = null, bool ascending = false)
+        public PagedResult<StoredMessage> GetAuditMessages(Endpoint endpoint, string searchQuery = null, string orderBy = null, bool ascending = false)
         {
-            var request = CreateMessagesRequest();
+            var request = CreateMessagesRequest(endpoint?.Name);
```
adamralph: This is now the only call to the `CreateMessagesRequest` method, so the parameter should be made mandatory there.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -97,76 +97,58 @@ public void CopyHeaders()
...
+
+                pagedResult.CurrentPage = pageIndex;
+                TryRebindMessageList(pagedResult);
+
+                SearchBar.IsVisible = true;
+                SearchBar.SetupPaging(pagedResult, link);
             }
+        }
 
+        public void RefreshMessages()
```
adamralph: This method used to fork, depending on whether the `selectedExplorerItem` was a `ServiceControlExplorerItem` or an `AuditEndpointExplorerItem`.

Is that no longer necessary?

src/ServiceInsight/MessageList/MessageListView.xaml.cs
```diff
@@ -56,7 +56,11 @@ void OnBeforeLayoutRefresh(object sender, CancelRoutedEventArgs e)
 
         void SortData(ColumnBase column, ListSortDirection order)
         {
-            Model.RefreshMessages(column.Tag as string, order == ListSortDirection.Ascending);
+            var orderBy = column.Tag as string;
+            var ascending = order == ListSortDirection.Ascending;
+
+            Model.ChangeSorting(orderBy, ascending);
+            Model.RefreshMessages();
```
SzymonPobiega: Change sorting is a local op, does not go to the server. The API now looks something like:
```
void Search(Endpoint endpoint, string searchQuery = null) //starts a brand new search from page 1
void NavigateToPage(string link, int pageIndex) //Continues a search using link returned from previous request
void RefreshMessages() //Starts a brand new search using values already present in the view model
```

src/ServiceInsight/MessageList/MessageListViewModel.cs
SzymonPobiega: This is now handled in `RefreshMessages`. Navigate uses  link that SC returned in a previous search request so it does not care about initial search params.

src/ServiceInsight/Search/SearchBarViewModel.cs
SzymonPobiega: That was probably some kind of a leftover. I'll remove it.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: SC can return either absolute or relative URLs, depending on the version :(

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: It would be quite complex I believe. Not sure how to best find the commas that are actually delimiters and not part of the URL.

SzymonPobiega:
@adamralph I've implemented your suggestions but I probably used `--ammend` in a wrong way. I'll fix that before merging.

src/ServiceInsight/MessageList/MessageListView.xaml.cs
adamralph: Given this is the only call to `ChangeSorting`, i.e. the first method is never called without calling the second, would it be better to encaspulate that behind a `RefreshMessages` overload which take `orderBy` and `ascending` parameters?

src/ServiceInsight/Search/SearchBarViewModel.cs
```diff
@@ -91,22 +90,27 @@ public void Search()
...
             CurrentPage = pagedResult.TotalCount > 0 ? pagedResult.CurrentPage : 0;
             TotalItemCount = pagedResult.TotalCount;
-
+            Result = pagedResult.Result;
+            NextLink = pagedResult.NextLink;
+            PrevLink = pagedResult.PrevLink;
+            FirstLink = pagedResult.FirstLink;
+            LastLink = pagedResult.LastLink;
+            PageSize = pagedResult.PageSize;
+            SelfLink = selfLink;
```
adamralph: This is the only reference to the `SelfLink` property, i.e. it is never read, only written to. Can it be removed?

If so, the `selfLink` parameter can also be removed from this method.

src/ServiceInsight/MessageList/MessageListView.xaml.cs
adamralph: _Continuing from https://github.com/Particular/ServiceInsight/pull/737/files#r133939640_

Given this is the only call to `ChangeSorting`, i.e. it is never called without then calling `RefreshMessages`, would it be better to encaspulate that behind a `RefreshMessages` overload with `orderBy` and `ascending` parameters?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -537,6 +557,7 @@ void LogError(IRestResponse response)
         static bool HasSucceeded(IRestResponse response) => successCodes.Any(x => response != null && x == response.StatusCode && response.ErrorException == null);
 
         static IEnumerable<HttpStatusCode> successCodes = new[] { HttpStatusCode.OK, HttpStatusCode.Accepted, HttpStatusCode.NotModified, HttpStatusCode.NoContent };
+        static Regex linkExpression = new Regex(@"<([^>]+)>;\s?rel=""(\w+)"",?\s?", RegexOptions.Compiled);
```
adamralph: _Continuing from https://github.com/Particular/ServiceInsight/pull/737/files#r133940544_

OK, let's rewind: can you explain how the regex works?

src/ServiceInsight/ServiceControl/IServiceControl.cs
```diff
@@ -34,5 +34,8 @@ public class ServiceControlHeaders
     {
         public const string ParticularVersion = "X-Particular-Version";
         public const string TotalCount = "Total-Count";
+        public const string Link = "Link";
+        public const string PageSize = "Page-Size";
+        public const string PageNumber = "Page-Number";
```
adamralph: We never use this, and instead we initialise the page number to 1 locally and then increment and decrement it locally when navigating. Is there any reason we don't read this from the response instead?

src/ServiceInsight/MessageList/MessageListViewModel.cs
adamralph: _Continuing from https://github.com/Particular/ServiceInsight/pull/737/files#r133939821_

I'm not really sure what you mean by "initial search params"", but that's probably because I don't know the difference between `ServiceControlExplorerItem` and `AuditEndpointExplorerItem`. 😁

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: yeah, sure. The first part `<([^>]+)>;` matches the url enclosed in `<` and `>` chars and a semicolon. Then there is an optional space and the rel part (`rel=""(\w+)""`). In the end there is an optional comma and space which separate one link from the other.

At first I tried to split on the commas to separate links and then on the semicolon to separate URL from rel part but commas can be also present in the URL. I would have to split only on the commas which are not between `<` and `>` which cannot be done declaratively but requires me to go through the string, char by char, counting `<` and `>`. 

I don't have a strong opinion and I am happy to rewrite it iteratively is you prefer this way.

src/ServiceInsight/ServiceControl/IServiceControl.cs
SzymonPobiega: This is not yet implemented in SC. I added this and then forgot to remove when I realized we need to first introduce it to SC to make the logic in SI simpler.

src/ServiceInsight/MessageList/MessageListViewModel.cs
SzymonPobiega: `ServiceControlExplorerItem` is present there if you don't select any endpoint. If you do, you get the `AuditEndpointExplorerItem` which contains the name of the endpoint to use to filter the audited messages.

SzymonPobiega:
@adamralph I force-pushed some changes you requested.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
adamralph: In your [previous comment](https://github.com/Particular/ServiceInsight/pull/737/files#r127168145), I see no mention of `<` and `>`. Could you please supply an example of the actual header value so I can understand better?

adamralph: Sorry, I made this comment on my last review at some point, but then got into a tangle so somehow it got removed:

_Continuning from https://github.com/Particular/ServiceInsight/pull/737/files?diff=unified#r134157914_

In your [previous comment](https://github.com/Particular/ServiceInsight/pull/737/files#r127168145), I see no mention of `<` and `>`. Could you please supply an example of the actual header value so I can understand better?

src/ServiceInsight/MessageList/MessageListView.xaml.cs
SzymonPobiega: My bad. Will fix it in a second.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
SzymonPobiega: Here's an example of the `Link` header:

```
<http://localhost:33333/api/messages?per_page=1&page=24>; rel="last", <http://localhost:33333/api/messages?per_page=1&page=2>; rel="next"
```

adamralph: Does SC really include non-encoded commas in the URL's? They should be encoded to `%2C`, then it would be a simple comma split. And semi-colons should be encoded to `%3B` allowing a further semi-colon split.

SzymonPobiega: Good question. I'll check that.

adamralph:
Marking as WIP until we confirm whether this fixes #736 or not.

SzymonPobiega:
@adamralph why? Fixing of #736 (if I can confirm it) was only a side effect of this PR.

adamralph:
@SzymonPobiega we want to know whether or not #736 should be closed when we merge this.

adamralph:
Current status:

- @SzymonPobiega to confirm whether SC URL encodes properly in the Link header, which will allow replacement of the regex with simple string splits
- @SzymonPobiega and @HEskandari to sync regarding #736, and specifically to determine if and how this PR affects that issue

adamralph:
Confirmed with @HEskandari and @SzymonPobiega that, after the merge of #745, this PR has no functional effect and is a pure refactoring.

Still marked as WIP, since there are currently some runtime errors caused by bad XAML -> code bindings.

HEskandari:
@SzymonPobiega Any updates on this? let me know if I can help.

SzymonPobiega:
@HEskandari I was away for a week of conferences/UGs. I hope to come back to this by the end of this week.

SzymonPobiega:
I rebased this. @HEskandari is there a way (other than by clicking) to make sure I didn't break more bindings?

HEskandari:
@SzymonPobiega no there is not, but since each feature is fairly isolated, it should be straightforward to test.

src/ServiceInsight/Search/SearchBarView.xaml
```diff
@@ -24,7 +24,7 @@
                 Padding="5"
                 HorizontalContentAlignment="Center"
                 VerticalContentAlignment="Center"
-                cal:Message.Attach="[Event Click]=[Action Search]">
+                cal:Message.Attach="[Event Click]=[Action RefreshResult]">
```
HEskandari: The search public method is still there. The RefreshResult is directly calling Search, so why not bind it directly to Search and remove the extra method?

SzymonPobiega: Because `CanSearch` is already defined and means something else. I had to use an explicit different name here so that I can have proper `Can*` property.

HEskandari: I still struggle to see the issue. If `CanSearch` is already defined that's fine too, since that evaluates the execution of both Search/RefreshResult (since that's what RefreshResult is calling anyway?)

Maybe jump on a call when you're feeling better and are back to work?

SzymonPobiega: @HEskandari the `CanSearch` is used in the search button as is defined as

```
public bool CanSearch => !WorkInProgress && !string.IsNullOrWhiteSpace(SearchQuery);
```

Which is different from `CanRefreshResult`:

```
public bool CanRefreshResult => !WorkInProgress;
```

The behavior of both buttons is the same but the conditions under which they are enabled are different. That's why I added the seemingly not required separate method.

SzymonPobiega:
I've rebased this on top of master.

danielmarbach:
Marked as WIP since we are investigating another approach

danielmarbach:
Removed WIP. As discussed with @WilliamBZA removed the PageCount and replaced the aboslute URI check with a more future proof one. Please review and merge. I will update the description with a few details around page count

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
WilliamBZA: I'm not able to make sense of the regex and would much rather prefer an iterative approach.

Is it sufficient to split on the first `,` that follows a `>;`? Or are there cases where there won't be any `>`'s?

Or what about splitting on '<'?

danielmarbach:
@WilliamBZA this is independent and can be merged and reviewed.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
danielmarbach: This is how the links look like

image

danielmarbach: Is the regex really a problem here? I mean string splitting can easily get as complex as understanding a regex.

SzymonPobiega: The splitting would be based on the context (we can't just split at `;` because it may be a part of a URL. I am glad that the SI maintainers accepted this solution. I would love to see this simplified.

WilliamBZA:
@Particular/serviceinsight-maintainers I've change the label to a feature as this has a material impact on the UX of the users.

We also need at least one more approval before squashing.

src/ServiceInsight.sln.DotSettings
```diff
@@ -562,6 +562,7 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateThisQualifierSettings/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002EJavaScript_002ECodeStyle_002ESettingsUpgrade_002EJsCodeFormatterSettingsUpgrader/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002EJavaScript_002ECodeStyle_002ESettingsUpgrade_002EJsParsFormattingSettingsUpgrader/@EntryIndexedValue">True</s:Boolean>
```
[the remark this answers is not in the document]
danielmarbach: For people using resharper this settings migration will always happen

src/ServiceInsight/Search/SearchBarView.xaml
```diff
@@ -69,7 +69,6 @@
             <Button Content="3" cal:Message.Attach="[Event Click]=[Action GoToPreviousPage]" />
             <TextBlock Text="Page" />
             <TextBlock Text="{Binding CurrentPage}" />
-            <TextBlock Text="{Binding PageCount, StringFormat=of {0}}" />
```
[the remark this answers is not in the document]
danielmarbach: See explanation in the PR description

image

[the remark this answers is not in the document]
adamralph: @danielmarbach thanks. I missed that. Is this a pre-existing problem? I.e. is the current method that SI uses to calculate the page flawed in the same (or a similar) way?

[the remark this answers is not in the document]
WilliamBZA: @adamralph The paging doesn't work with multiple instances of ServiceControl. There is no way to know what the "last" page is when connecting to more than 1 instance.

SzymonPobiega:
What is the main issue with this PR? Can we maybe jump on a call and talk through it?

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -104,76 +104,58 @@ public void CopyHeaders()
...
+                    searchQuery,
+                    sortColumn,
+                    sortOrderAscending);
 
                 if (pagedResult == null)
                 {
                     return;
                 }
 
+                pagedResult.CurrentPage = 1;
```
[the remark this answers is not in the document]
danielmarbach: I'm pretty sure this is actually required. You can be on any page and then when you search something you'll be forwarded to the first page by SI. I will verify though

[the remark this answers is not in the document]
danielmarbach: Checked it is not needed. Removing it

danielmarbach:
I removed the endpoint properties based on https://github.com/Particular/ServiceInsight/pull/737#discussion_r164468500
and also removed the setting of the current page.
The changes have been force pushed // cc @SzymonPobiega

danielmarbach:
Replaced by https://github.com/Particular/ServiceInsight/pull/773

Issue #733, opened 2017-06-27
Runtime error: Can't find property 'DxFts_10'


adamralph:
_From https://app.raygun.com/crashreporting/7gulz9/errors/1952419689/_

- **Version:** 1.7.0+14.Branch.master.Sha.3a8b8bdb65a6220e409691c88f82603bb9289351
- **Class name:** DevExpress.Data.Filtering.Exceptions.InvalidPropertyPathException
- **Message:** Can't find property 'DxFts_10'

<details>
<summary>Stacktrace</summary>
> DevExpress.Data.Filtering.Helpers.EvaluatorCriteriaValidator.Visit(OperandProperty theOperand):86
> DevExpress.Data.Filtering.Helpers.EvaluatorCriteriaValidator.Validate(IList operands):28
> DevExpress.Data.Filtering.Helpers.EvaluatorCriteriaValidator.Visit(FunctionOperator theOperator):12
> DevExpress.Data.Filtering.Helpers.EvaluatorCriteriaValidator.Validate(IList operands):28
> DevExpress.Data.Filtering.Helpers.EvaluatorCriteriaValidator.Visit(GroupOperator theOperator):12
> DevExpress.Data.Filtering.Helpers.ExpressionEvaluator..ctor(PropertyDescriptorCollection properties, CriteriaOperator criteria, Boolean caseSensitive, Boolean doCreateEvauluatorCore):33
> DevExpress.Xpf.PropertyGrid.Internal.DataViewBase.get_ExpressionEvaluator():23
> DevExpress.Xpf.PropertyGrid.Internal.DataViewBase.IsVisible(RowHandle handle):0
> DevExpress.Xpf.PropertyGrid.Internal.DataViewBase+<GetFilteredChildrenInternal>d__1d.MoveNext():99
> System.Linq.Enumerable.Any[TSource](IEnumerable`1 source):21
> DevExpress.Xpf.PropertyGrid.Internal.DataViewBase.CanExpand(RowHandle handle):53
> DevExpress.Xpf.PropertyGrid.Internal.RowDataGenerator.PrepareRowDataForHandleOverride(RowDataBase data, RowHandle handle, RowDataProperties properties):78
> DevExpress.Xpf.PropertyGrid.Internal.RowDataGenerator.InitializeViewItemsSource(PropertyGridView view):34
> DevExpress.Xpf.PropertyGrid.Internal.RowDataGenerator.DoActionAndSaveFocus(Action action):123
[...]

PR #740, opened 2017-08-11
Fix property grid search exception

WilliamBZA:
Connects to https://github.com/Particular/ServiceInsight/issues/733

* Updates to the latest version of DevExpress that we are licensed for
* Includes a value converter and behavior to wrap search strings in quotes - this prevents the exception from occurring


src/ServiceInsight/ValueConverters/SearchStringConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+                str = str.Substring(1, str.Length - 2);
+            }
+
+            return str;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return string.Format("\"{0}\"", value);
+        }
```
adamralph: This could be more for my own education than anything else, but looking purely at the method signatures, I would naively write these methods like so:

```C#
public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
{
    if (targetType != typeof(string))
    {
        throw new InvalidOperationException($"Cannot convert value to target type '{targetType}'.");
    }

    if (value == null)
    {
        return null;
    }

    var stringValue = value as string;

    if (stringValue == null)
    {
        throw new InvalidOperationException($"The value is not a string.");
    }

    return stringValue.StartsWith("\"", true, culture) && stringValue.EndsWith("\"", true, culture)
        ? stringValue.Substring(1, stringValue.Length - 2)
        : stringValue;
}

public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) =>
    string.Format(culture, "\"{0}\"", value);
```

However, I don't know what assumptions can be made wrt the WPF execution context, and what does or does not need to be done. Things to note about the above snippet:

- Respecting the `targetType` argument and throwing if it's not a `string`
- Short circuiting the case where `value` is `null`.
- Avoiding the string allocation by safe casting `value` to a `string`.
- Throwing if value is not a `string`
- using the supplied `culture` argument where appropriate (in both methods)

WilliamBZA: I'm not a real fan of the exceptions. The converters are invoked by the databinding pipeline so any exceptions are [treated as application exceptions](https://msdn.microsoft.com/en-us/library/system.windows.data.ivalueconverter.convert.aspx?f=255&mspperror=-2147217396#Anchor_1) (could crash?)

What about:

```CSharp
public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
{
    if (targetType != typeof(string))
    {
        return DependencyProperty.UnsetValue;
    }

    var stringValue = value as string;
    if (stringValue == null)
    {
        return DependencyProperty.UnsetValue;
    }

    return stringValue.StartsWith("\"", true, culture) && stringValue.EndsWith("\"", true, culture)
        ? stringValue.Substring(1, stringValue.Length - 2)
        : stringValue;
}
```

HEskandari: You're both right! You could use exceptions in a converter and they show up as warnings, but the right approach is to use the `DependencyProperty.UnsetValue` as @WilliamBZA highlighted.

adamralph: @WilliamBZA your suggestion LGTM.

WilliamBZA:
Updated based on feedback and rebased on master.

I think this is good to go now

src/ServiceInsight/Framework/Behaviors/ComboBoxEditValuePatcherBehavior.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace ServiceInsight.Framework.Behaviors
+{
+    using System.Windows.Data;
+    using DevExpress.Mvvm.UI.Interactivity;
+    using DevExpress.Xpf.Editors;
+    using ServiceInsight.ValueConverters;
+
+    public class ComboBoxEditValuePatcherBehavior : Behavior<ComboBoxEdit>
```
HEskandari: I suppose this is the 'fix'? Should we put a comment there or somehow make it more obvious that this is not a generic behavior? Maybe a link to the bug report?

Issue #738, opened 2017-07-12
SagaWindowViewModel.FetchOrderedSagaData - NullReferenceException


danielHalan:
Hi, 

Following Error occurs on ServiceInsight startup, causing it to Terminate.

`System.NullReferenceException: Object reference not set to an instance of an object.
   at ServiceInsight.Saga.SagaWindowViewModel.<>c__DisplayClass31_0.<FetchOrderedSagaData>b__3(SagaUpdate update) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 222
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at ServiceInsight.Saga.SagaWindowViewModel.FetchOrderedSagaData(Guid sagaId) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 221
   at ServiceInsight.Saga.SagaWindowViewModel.RefreshSaga(SagaInfo originatingSaga) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 170
   at ServiceInsight.Saga.SagaWindowViewModel.RefreshSaga(StoredMessage message) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 152
   at ServiceInsight.Saga.SagaWindowViewModel.Handle(SelectedMessageChanged event) in C:\BuildAgent\work\7540559e45164f93\src\ServiceInsight\Saga\SagaWindowViewModel.cs:line 103`

HEskandari:
Thanks for reporting it @danielHalan. I'm just wondering which version of SI were you using? There seems to be an issue with the Saga data that is potentially expired? Do you have this symptom on other instances of ServiceControl? 

Would you be able to securely share the database with us so we can reproduce it? If yes, I can email you the details on how.

danielHalan:
SI v1.7.1, you can contact me for diagnostics details, using my github handle _at_ gmail

HEskandari:
@danielHalan I had a look through the code and we are checking the data for being null.  Since you couldn't reproduce the issue either, should we close this for now? (we can always reopen it if you encounter it again).

HEskandari:
Keeping this issue open for now as I have received another report of this exact problem.

[...]

PR #742, opened 2017-08-16
No longer assume object is properly hydrated

WilliamBZA:
Connects to https://github.com/Particular/ServiceInsight/issues/738

Doesn't assume that the InitiatingMessage object will exist. This object doesn't exist in the case where a saga has > 200 state changes where we inject a `missing_data` response to indicate that we're truncating the results.


adamralph:
Marking as WIP until a test is added (assuming it's practical).

HEskandari:
Removed the WIP as the test is added.

PR #745, opened 2017-08-28
Selects the root node in the endpoint explorer after connecting to SC

### LINKED ISSUE #736: Connecting to ServiceControl does not trigger a refresh of the message list when the endpoint explorer does not have focus

HEskandari:
Resolves #736


adamralph:
Confirmed today with @HEskandari and @SzymonPobiega that this fixes #736.

PR #730, opened 2017-06-09
[WIP] Added three new columns to Messages window

HEskandari:
Connects to #723.
~Closes #734.~
Connects to #735.

PR #746, opened 2017-08-29
Column chooser for Messages List window and persistence of the layout

### LINKED ISSUE #734: Add a column chooser
Column selection to be persisted across sessions.

## Plan of attack

- [x] Add column chooser
- [x] Persist across sessions

HEskandari:
Closes #734

### DIFF
--- a/src/ServiceInsight/ExtensionMethods/LayoutExtensions.cs
+++ b/src/ServiceInsight/ExtensionMethods/LayoutExtensions.cs
@@ -1,0 +1,76 @@
+﻿namespace ServiceInsight.ExtensionMethods
+{
+    using System;
+    using System.IO;
+    using Anotar.Serilog;
+    using DevExpress.Xpf.Bars;
+    using DevExpress.Xpf.Docking;
+    using DevExpress.Xpf.Grid;
+
+    public static class LayoutExtensions
+    {
+        public static string GetLayout(this BarManager barManager)
+        {
+            return TrySaveLayout(barManager);
+        }
+
+        public static string GetLayout(this DockLayoutManager dockLayoutManager)
+        {
+            return TrySaveLayout(dockLayoutManager);
+        }
+
+        public static string GetLayout(this GridControl gridControl)
+        {
+            return TrySaveLayout(gridControl);
+        }
+
+        public static void RestoreLayout(this BarManager barManager, Stream stream)
+        {
+            TryRestoreLayout(barManager, stream);
+        }
+
+        public static void RestoreLayout(this DockLayoutManager dockLayoutManager, Stream stream)
+        {
+            TryRestoreLayout(dockLayoutManager, stream);
+        }
+
+        public static void RestoreLayout(this GridControl gridControl, Stream stream)
+        {
+            TryRestoreLayout(gridControl, stream);
+        }
+
+        static string TrySaveLayout(dynamic control) //Lack of common interface :(
+        {
+            try
+            {
+                using (var ms = new MemoryStream())
+                {
+                    control.SaveLayoutToStream(ms);
+                    return ms.GetAsString();
+                }
+            }
+            catch (Exception ex)
+            {
+                LogTo.Information(ex, "Failed to save the layout, reason is: {ex}", ex);
+                return null;
+            }
+        }
+
+        static void TryRestoreLayout(dynamic control, Stream layout)
+        {
+            if (layout == null)
+            {
+                return;
+            }
+
+            try
+            {
+                control.RestoreLayoutFromStream(layout);
+            }
+            catch (Exception ex)
+            {
+                LogTo.Information(ex, "Failed to restore layout, reason is: {ex}", ex);
+            }
+        }
+    }
+}
--- a/src/ServiceInsight/MessageList/MessageListView.xaml.cs
+++ b/src/ServiceInsight/MessageList/MessageListView.xaml.cs
@@ -7,8 +7,12 @@
     using System.Windows.Input;
     using DevExpress.Xpf.Core;
     using DevExpress.Xpf.Grid;
+    using ServiceInsight.ExtensionMethods;
+    using ServiceInsight.Framework.Settings;
+    using ServiceInsight.Settings;
+    using ServiceInsight.Shell;
 
-    public interface IMessageListView
+    public interface IMessageListView : IPersistableLayout
     {
         void BeginDataUpdate();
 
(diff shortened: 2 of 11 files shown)


HEskandari:
I have separated the column chooser bits from PR #730. Please review.

src/ServiceInsight/ExtensionMethods/LayoutExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                using (var ms = new MemoryStream())
+                {
+                    control.SaveLayoutToStream(ms);
+                    return ms.GetAsString();
+                }
+            }
+            catch (Exception ex)
+            {
+                LogTo.Information(ex, "Failed to save the layout, reason is: {ex}", ex);
```
WilliamBZA: Is Info the right log level here? It doesn't cause the app to crash, but perhaps debug is appropriate.

adamralph: Given this is a non-critical failure, I'd say WARN is the appropriate level.

src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -286,5 +293,15 @@ public void BringIntoView(StoredMessage msg)
...
+
+        public void OnSavePartLayout()
+        {
+            view.OnSaveLayout(settingsProvider);
+        }
+
+        void RestoreLayout()
+        {
+            view.OnRestoreLayout(settingsProvider);
+        }
```
adamralph: I see that `RestoreLayout` is called from `OnViewAttached`. Is there an equivalent method that could call `OnSavePartLayout`, to remove the need for the `IPersistPartLayout` interface, and its explicit call from `ShellViewModel`?

src/ServiceInsight/MessageList/MessageListView.xaml.cs
```diff
@@ -102,5 +106,25 @@ void OnGridClicked(object sender, MouseButtonEventArgs e)
...
+        {
+            var layoutSetting = settingsProvider.GetSettings<MessageListSettings>();
+            grid.RestoreLayout(layoutSetting.GridLayout.GetAsStream());
+        }
+
+        public void OnResetLayout(ISettingsProvider settingsProvider)
+        {
+            var layoutSettings = settingsProvider.GetSettings<MessageListSettings>();
+            layoutSettings.GridLayout = null;
+            settingsProvider.SaveSettings(layoutSettings);
```
adamralph: Why does nothing happen with `grid` in this method?

src/ServiceInsight/ExtensionMethods/LayoutExtensions.cs
adamralph: After further review, I see that these log messages are pre-existing and have been moved from `Shell/Shell.xaml.cs`, i.e. the saving and restoration of the new settings for the column chooser are following the pattern of that of the existing settings for the layout of shell. For that reason I think we should keep this as-is for this PR. If we want to change the log level and messages (I think they also need some work) for problems in persisting all settings, I believe we should do that a a separate PR.

src/ServiceInsight/Shell/ShellViewModel.cs
```diff
@@ -132,15 +133,23 @@ protected override void OnViewAttached(object view, object context)
 
         protected override void OnDeactivate(bool close)
         {
-            base.OnDeactivate(close);
             refreshTimer.Stop();
             SaveLayout();
+            base.OnDeactivate(close);
```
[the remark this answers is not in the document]
HEskandari: the base should be called last, as it deactivates the screen and sets a few other flags, not allowing SaveLayout to properly run.

src/ServiceInsight/Shell/ShellViewModel.cs
```diff
@@ -132,15 +133,23 @@ protected override void OnViewAttached(object view, object context)
...
         {
             if (!comandLineArgParser.ParsedOptions.ResetLayout)
             {
+                var persistableLayouts = Items.OfType<IPersistPartLayout>();
+                {
+                    foreach (var screen in persistableLayouts)
+                    {
+                        screen.OnSavePartLayout();
+                    }
+                }
```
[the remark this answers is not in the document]
HEskandari: Here's how it works:

Items is a property of all the parts/sub-screens in the shell. Some of the parts/sub-screens may implement the interface `IPersistPartLayout` that says they need to serialize their layout upon exit. 

So to answer: if we move `Item.OfType<IPersistPartLayout>()` to a well-named method, would it become simpler?

src/ServiceInsight/MessageList/MessageListViewModel.cs
HEskandari: The ViewAttach stuff comes from `IViewAware` interface of caliburn. It doesn't have anything for the view is detached.

The interface `IPersistPartLayout` still would be required though, as it is an opt-in interface.

src/ServiceInsight/MessageList/MessageListView.xaml.cs
HEskandari: This is the Reset method, which would null-out the stored layout in the settings (so it falls back to the default layout). This setting - generally speaking, all settings - become effective on the next restart (we have a UI hint for it as well).

src/ServiceInsight/ExtensionMethods/LayoutExtensions.cs
HEskandari: 👍  on doing it separately.

src/ServiceInsight/Shell/ShellViewModel.cs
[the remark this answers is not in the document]
adamralph: @HEskandari I believe I understand how it works, and the implementation looks fine.

All I'm suggesting is replacing this block:

```C#
var persistableLayouts = Items.OfType<IPersistPartLayout>();
{
    foreach (var screen in persistableLayouts)
    {
        screen.OnSavePartLayout();
    }
}
```

with this block:

```C#
foreach (var item in Items.OfType<IPersistPartLayout>())
{
    item.OnSavePartLayout();
}
```

It's functionally identical, and a pure refactoring. Would that work?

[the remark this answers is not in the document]
HEskandari: @adamralph ah sure :) that works just fine. I'll update.

HEskandari:
@adamralph The PR is update (and rebased). I'll squash the commits after the reviews are done.

HEskandari:
@adamralph all done.

PR #750, opened 2017-09-01
Delays the binding to selected message property

### LINKED ISSUE #721: Listview should use delayed event handling
# Problem

When browsing through the items in the listview SI is immediately loading the results and locks the listview.

# Solution

It should use delayed event handling that only loads an item if it has been selected for more then for example 50 milliseconds. This allows to quickly scroll through the items with the keyboard without actually loading each item.

HEskandari:
Resolves #721


adamralph:
@ramonsmits as the original requester of this change, would you mind trying out this branch to see if it solves the problem adequately? Note that the current delay is 100ms, but we've also been experimenting with 50ms, to see what's better. Please could you test both values and let us know what you prefer.

HEskandari:
@adamralph Changed the delay to 50ms. Please merge if there's nothing else.

PR #753, opened 2017-09-08
Information on how to resolve missing System.Windows.Interactivity assembly

ramonsmits:


WilliamBZA:
I am not actually sure of the value here. If the prerequisites are installed, the error doesn't occur.  The clarification only helps if someone is reading this set of instructions. If they're reading this set of instructions, they will install the prerequisites and won't get the error.

adamralph:
@WilliamBZA my initial reaction was the same, before I convinced myself that this is a "belt and braces" approach worth doing.

E.g. someone might clone and build without reading the doc, see the error and then Google "ServiceInsight System.Windows.Interactivity" (or search the repo directly) and get no hits. With this in place, they'd get a hit on the troubleshooting point.

What do you think?

HEskandari:
I'd say merge it. This will connect the error message which has no connection to Blend SDK to the prerequisite.

PR #754, opened 2017-09-11
Escape key binding convention for all the windows created via WindowManager

### LINKED ISSUE #711: Close dialogs with escape key

HEskandari:
Resolves #711

### DIFF
--- a/src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs
+++ b/src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs
@@ -1,0 +1,27 @@
+﻿namespace ServiceInsight.Framework.Commands
+{
+    using System;
+    using System.Windows;
+    using System.Windows.Input;
+
+    public class CloseWindowCommand : ICommand
+    {
+        readonly Window target;
+
+        public CloseWindowCommand(Window target) => this.target = target;
+
+        public bool CanExecute(object parameter) => target != null;
+
+        public void Execute(object parameter)
+        {
+            if (CanExecute(parameter))
+            {
+                target.Close();
+            }
+        }
+
+#pragma warning disable CS0067
+        public event EventHandler CanExecuteChanged;
+#pragma warning restore CS0067
+    }
+}
--- a/src/ServiceInsight/Framework/UI/ScreenManager/WindowManagerEx.cs
+++ b/src/ServiceInsight/Framework/UI/ScreenManager/WindowManagerEx.cs
@@ -3,7 +3,9 @@
     using System.Collections.Generic;
     using System.Windows;
     using System.Windows.Forms;
+    using System.Windows.Input;
     using Caliburn.Micro;
+    using ServiceInsight.Framework.Commands;
 
     public class WindowManagerEx : WindowManager, IWindowManagerEx
     {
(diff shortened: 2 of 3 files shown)


src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+        public void Execute(object parameter)
+        {
+            if (CanExecute(parameter))
+            {
+                target.Close();
+            }
+        }
+
+        #pragma warning disable CS0067
+        public event EventHandler CanExecuteChanged;
```
WilliamBZA: We should really re-enable the warning again.

```CSharp
#pragma warning disable CS0067
public event EventHandler CanExecuteChanged;
#pragma warning restore CS0067

Comment (HEskandari) [reply src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs]: Thanks @WilliamBZA forgot about that :)

Comment (adamralph) [reply src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs]: I saw that too but because it was at the end of the class I let it go. However, 👍 for correctness. I'm ashamed of myself. 😜

File: src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+        #pragma warning restore CS0067
```
HEskandari: @adamralph @WilliamBZA this is now addressed.

src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+    using System.Windows.Input;
+
+    public class CloseWindowCommand : ICommand
+    {
+        readonly Window target;
+
+        public CloseWindowCommand(Window target)
+        {
+            this.target = target;
+        }
```
adamralph: This constructor can be expression-bodied.

src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public bool CanExecute(object parameter)
+        {
+            return target != null;
+        }
```
adamralph: This method can be expression-bodied.

src/ServiceInsight/Framework/Commands/CloseWindowCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public void Execute(object parameter)
+        {
+            if (CanExecute(parameter))
+            {
+                target.Close();
+            }
+        }
+
+        #pragma warning disable CS0067
```
adamralph: Minor nit-pick, but the two pragma line should not be indented. An auto-format will remove the indents, so it's probably better to get rid of them now.

src/ServiceInsight/Framework/UI/ScreenManager/WindowManagerEx.cs
```diff
@@ -91,6 +102,12 @@ public MessageBoxResult ShowMessageBox(string message, string caption = "", Mess
             return ShowDialog(model, null, settings);
         }
 
+        static void ApplyExitOnEscape(Window window)
+        {
+            var escapeBinding = new KeyBinding(new CloseWindowCommand(window), new KeyGesture(Key.Escape));
+            window.InputBindings.Add(escapeBinding);
+        }
```
adamralph: Can we inline this method? It seems overkill to have these two lines in a private method with only one caller.

I'd also consider inlining the `escapeBinding` variable.

HEskandari: Updated the PR and incorporated the feedbacks

PR #756, opened 2017-09-26
Bypassing self-signed certificate validation

### LINKED ISSUE #717: Configuration option to skip certificate validation for connection to ServiceControl
This is a spin-off of the discussion in https://github.com/Particular/ServiceInsight/issues/604.

This allows, for example, connections to ServiceControl running with a self-signed certificate.

We will allow this via an app config setting, which address the problem whether ServiceInsight is launched by the user directly, or via ServicePulse.

## Plan of attack

- [ ] Make changes - #756
- [ ] Update docs - https://github.com/Particular/docs.particular.net/pull/3273

HEskandari:
Resolves #717

### DIFF
--- a/src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
+++ b/src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
@@ -1,0 +1,29 @@
+﻿namespace ServiceInsight.Framework.Settings
+{
+    using System;
+    using System.Configuration;
+
+    public static class ApplicationConfiguration
+    {
+        public static bool SkipCertificateValidation { get; private set; }
+
+        public static void Initialize()
+        {
+            SkipCertificateValidation = GetSkipCertificateValidation();
+        }
+
+        private static bool GetSkipCertificateValidation()
+        {
+            var skipCertificateValidationValue = ConfigurationManager.AppSettings["SkipCertificateValidation"];
+
+            try
+            {
+                return Convert.ToBoolean(skipCertificateValidationValue);
+            }
+            catch (Exception ex)
+            {
+                throw new ConfigurationErrorsException("Application setting 'SkipCertificateValidation' cannot be converted to type Boolean.", ex);
+            }
+        }
+    }
+}
--- a/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
+++ b/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
@@ -41,6 +41,11 @@
         IEventAggregator eventAggregator;
         ProfilerSettings settings;
 
+        static DefaultServiceControl()
+        {
+            ServicePointManager.ServerCertificateValidationCallback = (sender, certificate, chain, errors) => ApplicationConfiguration.SkipCertificateValidation;
+        }
+
         public DefaultServiceControl(
             ServiceControlConnectionProvider connection,
             IEventAggregator eventAggregator,
--- a/src/ServiceInsight/app.config
+++ b/src/ServiceInsight/app.config
@@ -36,4 +36,7 @@
       </dependentAssembly>
     </assemblyBinding>
   </runtime>
+  <appSettings>
+    <add key="SkipCertificateValidation" value="False" />
+  </appSettings>
 </configuration>
--- a/src/ServiceInsight/App.xaml.cs
+++ b/src/ServiceInsight/App.xaml.cs
@@ -6,6 +6,7 @@
     using Anotar.Serilog;
     using DevExpress.Xpf.Core;
     using Framework.Logging;
+    using ServiceInsight.Framework.Settings;
     using Shell;
 
     public interface IAppCommands
@@ -53,6 +54,7 @@
         {
             LogTo.Information("Starting the application...");
             DXSplashScreen.Show(o => AboutView.AsSplashScreen(), null, null, null);
+            ApplicationConfiguration.Initialize();
             base.OnStartup(e);
             LogTo.Information("Application startup finished.");
         }
--- a/src/ServiceInsight/ServiceInsight.csproj
+++ b/src/ServiceInsight/ServiceInsight.csproj
@@ -262,6 +262,7 @@
     <Compile Include="Framework\Behaviors\CommandParameterHelper.cs" />
     <Compile Include="Framework\Commands\DelegateCommand.cs" />
     <Compile Include="Framework\Commands\ObservableCommand.cs" />
+    <Compile Include="Framework\Settings\ApplicationConfiguration.cs" />
     <Compile Include="Framework\WorkNotifier.cs" />
     <Compile Include="Framework\Commands\BaseCommand.cs" />
     <Compile Include="Framework\Commands\ChangeSelectedMessageCommand.cs" />


HEskandari:
@adamralph I stand corrected, the callback actually is called even when connecting to a valid HTTPS endpoint. I'll rename the config entry to `SkipCertificateValidation`

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace ServiceInsight.Framework.Settings
+{
+    using System;
+    using System.Configuration;
+
+    public class ApplicationConfiguration
+    {
+        public static bool SkipCertificateValidation => Convert.ToBoolean(ConfigurationManager.AppSettings["SkipCertificateValidation"]);
```
adamralph: If the string is not a valid boolean representation, we'll get a `System.FormatException` with the message "String was not recognized as a valid Boolean.". While the stack trace could indicate what the problem is, I guess this could be subject to optimisation and may not be guaranteed to do so. I think it would be better to catch any conversion exception and wrap it in an exception which calls out the name of the configuration item.

adamralph: I'm not sure a static member is the best way to do this. I'd prefer the app entry point (Program.cs) to reach into app.config and set this value at application start up.

src/ServiceInsight/Properties/Settings.settings
```diff
@@ -1,7 +0,0 @@
-﻿<?xml version='1.0' encoding='utf-8'?>
-<SettingsFile xmlns="uri:settings" CurrentProfile="(Default)">
-  <Profiles>
-    <Profile Name="(Default)" />
-  </Profiles>
-  <Settings />
-</SettingsFile>
```
adamralph: This file (and Settings.Designer.cs and the related elements in the csproj) should be untouched by this PR. If they are redundant then we can remove them in another PR (or at least, as a separate clean up commit in this PR).

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -41,6 +41,11 @@ public class DefaultServiceControl : IServiceControl
         IEventAggregator eventAggregator;
         ProfilerSettings settings;
 
+        static DefaultServiceControl()
+        {
+            ServicePointManager.ServerCertificateValidationCallback = (sender, certificate, chain, errors) => ApplicationConfiguration.SkipCertificateValidation;
```
adamralph: The problem with this is that every single time the certificate is checked, it will call back into app.config. As I said above, I believe Program.cs should read the app.config value and set this property at app start up.

src/ServiceInsight/app.config
```diff
@@ -36,4 +36,7 @@
       </dependentAssembly>
     </assemblyBinding>
   </runtime>
+	<appSettings>
+		<add key="SkipCertificateValidation" value="False" />
+	</appSettings>
```
adamralph: These lines are using tabs. They should be converted to spaces.

HEskandari: Done

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
HEskandari: Agreed. There's no Program.cs so moved out to AppBootstrapper that executes once on app_startup.

src/ServiceInsight/Properties/Settings.settings
HEskandari: Brought them back. There are a few other issues I discovered along the way. I'll submit a refactoring PR to get them fixed.

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
HEskandari: Done

HEskandari: I have changed the way the values are parsed. Now you get a better exception details/message. The exception will be thrown if the value can not be parsed or the key is missin]g altogether, which I find to cause fewer surprises.

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        private static T GetValue<T>(string key)
+        {
+            try
+            {
+                return (T)Convert.ChangeType(ConfigurationManager.AppSettings[key], typeof(T));
+            }
+            catch (Exception ex)
+            {
+                throw new ApplicationException($"The value of '{key}' in app.config is not a valid {typeof(T).Name} type.", ex);
```
adamralph: I see a few problems with this exception message:

- The value of a given app setting is not a "type", but rather a string representation of an instance of a type.
- The user will never see a file named "app.config".
- The value may be missing entirely, as another reason for the conversion to fail.
- We should really use the full type name.

Perhaps:

```C#
$"Application setting '{key}' is either missing or cannot be converted to type {typeof(T)}."
```

However, there is also another problem here. If `T` is a nullable type, e.g. `string`, the behaviour will be different to that of a non-nullable type, e.g. `bool`. In the first case, a null will be returned. In the latter, an exception will be thrown. Perhaps we should first retrieve the value and check if it is null, and throw a separate exception for that, and only if it is not null, should we then proceed to the conversion.

src/ServiceInsight/Startup/AppBootstrapper.cs
```diff
@@ -27,6 +28,8 @@ public AppBootstrapper()
 
         protected override void Configure()
         {
+            ApplicationConfiguration.Initialize();
```
adamralph: Would this be better placed in `OnStartup`?

The docs for `Configure` say:

> Override to configure the framework and setup your IoC container.

But this method neither configures the framework nor sets up the IoC container.

The docs for `OnStartup` are not very specific:

> Override this to add custom behavior to execute after the application starts.

However, `OnStartup` receives a `StartupEventArgs` which has an `Args` property, and the docs for that say:

> Gets command line arguments that were passed to the application from either the command prompt or the desktop

This suggests that `OnStartup` is the dual of `Program.Main` so I think it makes sense to read the app config here.

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
HEskandari: I'd say over-thinking it to solve all possible scenarios would be premature optimization. It has taken us quite a few years without needing the app.config file at all. I was thinking to go with a more specific/simpler version to target the problem at hand (e.g. a non-generic method that deals with a bool).

src/ServiceInsight/Startup/AppBootstrapper.cs
HEskandari: I didn't want to put too many things into the app startup, since we need to:

1. Start logging, wire exception handler
2. Show the splash screen.

If we do this before step 1, it might blow up with an unhandled exception. If we keep adding stuff before step 2, the splash screen will take long to appear which beats its purpose.

In this case, I think since there's little to do, between step 1 and 2 should be okay. That means it will be moved to the end of the App constructor. Should I move it there?

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
HEskandari: BTW, the problem of having a typed access to app.config was already solved by Settings files, aren't we re-creating the wheel just for consistency sake?

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -9,18 +9,18 @@ public class ApplicationConfiguration
...
         {
             try
             {
-                return (T)Convert.ChangeType(ConfigurationManager.AppSettings[key], typeof(T));
+                return Convert.ToBoolean(ConfigurationManager.AppSettings["SkipCertificateValidation"]);
             }
             catch (Exception ex)
             {
-                throw new ApplicationException($"The value of '{key}' in app.config is not a valid {typeof(T).Name} type.", ex);
+                throw new ApplicationException("Application setting 'SkipCertificateValidation' is either missing or cannot be converted to type boolean.", ex);
```
adamralph: Nitpick: the casing should be "Boolean".

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace ServiceInsight.Framework.Settings
+{
+    using System;
+    using System.Configuration;
+
+    public class ApplicationConfiguration
```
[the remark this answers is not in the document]
adamralph: This class should be `static`.

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        private static bool GetSkipCertificateValidation()
+        {
+            try
+            {
+                return Convert.ToBoolean(ConfigurationManager.AppSettings["SkipCertificateValidation"]);
+            }
+            catch (Exception ex)
+            {
+                throw new ApplicationException("Application setting 'SkipCertificateValidation' is either missing or cannot be converted to type boolean.", ex);
```
adamralph: `ApplicationException` is effectively deprecated. See https://docs.microsoft.com/en-gb/dotnet/api/system.applicationexception?view=netframework-4.7 (emphasis mine):

> ⚠️ Important
> You should derive custom exceptions from the Exception class rather than the ApplicationException class. **You should not throw an ApplicationException exception in your code**, and you should not catch an ApplicationException exception unless you intend to re-throw the original exception.

I think the correct type to use could be [`ConfigurationErrorsException`](https://docs.microsoft.com/en-gb/dotnet/api/system.configuration.configurationerrorsexception?view=netframework-4.7) (using the same `string`, `Exception` constructor).

adamralph: Nitpick: the casing in the exception message should be "Boolean".

src/ServiceInsight/Properties/Settings.settings
```diff
@@ -1,7 +1,9 @@
...
-  <Profiles>
-    <Profile Name="(Default)" />
-  </Profiles>
-  <Settings />
+<SettingsFile xmlns="http://schemas.microsoft.com/VisualStudio/2004/01/settings" CurrentProfile="(Default)" GeneratedClassNamespace="ServiceInsight.Properties" GeneratedClassName="Settings">
+  <Profiles />
+  <Settings>
+    <Setting Name="AllowSelfSignedCertificates" Type="System.Boolean" Scope="Application">
+      <Value Profile="(Default)">False</Value>
+    </Setting>
```
adamralph: As discussed, we decided not to use the settings file for this value. This entire file (and Settings.Designer.cs) should be reverted.

HEskandari:
@serviceinsight-maintainers the requested changes are made

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+        public static void Initialize()
+        {
+            SkipCertificateValidation = GetSkipCertificateValidation();
+        }
+
+        private static bool GetSkipCertificateValidation()
+        {
+            try
+            {
+                return Convert.ToBoolean(ConfigurationManager.AppSettings["SkipCertificateValidation"]);
```
WilliamBZA: what about

```CSharp
return Convert.ToBoolean(ConfigurationManager.AppSettings["SkipCertificateValidation"] ?? "false");
```

to cover for a default "false" value? That way we can change the exception to only be a valid format check instead of requiring the value to exist in the config file?

HEskandari: We're shipping the app.config with the default value. Why have two?

WilliamBZA: Just feels weird to not have a default. I have no strong preference one way or the other. Happy with the changes as is.

HEskandari: But there _is_ a default, it resides in the config file. I think having a fallback value, would make it confusing for the user as there's no indication we're using that value.

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+        public static void Initialize()
+        {
+            SkipCertificateValidation = GetSkipCertificateValidation();
+        }
+
+        private static bool GetSkipCertificateValidation()
+        {
+            try
+            {
+                return Convert.ToBoolean(ConfigurationManager.AppSettings["SkipCertificateValidation"] ?? "false");
```
adamralph: Perhaps I'm getting too far into an edge case, but could there be a situation where some other exception is thrown, and we disguise it as the value not being a valid Boolean? E.g. a problem accessing the config file itself? To be safe, should we only try...catch `Convert.ToBoolean()` and not `ConfigurationManager.AppSettings[]`?

adamralph: I looked into `Convert.ToBoolean()` and it turns out that it already converts `null` to `false`:

image

Indeed, this is the [only difference between `Convert.ToBoolean()` and `bool.Parse()`](https://stackoverflow.com/a/7032036/49241).

This makes `?? "false"` redundant. Shall we remove it?

WilliamBZA: The only bug free code is deleted code. :+1: to remove

HEskandari:
Rolled back :)

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
HEskandari: Rolled back!

HEskandari: @adamralph in which case, the error message + inner exception should be clear enough to find that out, right?

WilliamBZA: Exception message doesn't match. If the value is missing the default will be false. The `is either missing or` doesn't make sense.

HEskandari: @WilliamBZA I've already rolled that change back though. See the revert commit below.

src/ServiceInsight/Framework/Settings/ApplicationConfiguration.cs
WilliamBZA: Exception is still

`throw new ConfigurationErrorsException("Application setting 'SkipCertificateValidation' is either missing or cannot be converted to type Boolean.", ex);`

It will never throw if it's missing as per Adam's investigation

WilliamBZA:
Exception updated. @Particular/serviceinsight-maintainers are you happy for me to squash?

HEskandari:
@WilliamBZA :+1:

WilliamBZA:
I've made the change @adamralph. Ready for squash now?

PR #759, opened 2017-11-01
Update signing thumbprint

kbaley:
The signing certificate is about to expire on December 22. A replacement has been installed on all the build agents. This is the thumbprint for the new one


HEskandari:
Thanks @kbaley

kbaley:
@Particular/serviceinsight-maintainers Not familiar with the deployment process here. Do I need to do anything to release a new version of the installer?

WilliamBZA:
I believe it will be created when we do the next release. I'm not sure how it gets brought into the Platform Installer though, I expect @HEskandari would know.

kbaley:
@HEskandari answered on [ServicePulse](https://github.com/Particular/ServicePulse/pull/543#issuecomment-342039091)

PR #762, opened 2017-11-04
Use a code default instead of config file

WilliamBZA:
It seems that when upgrading the config file is not replaced, so we could be in a situation where users upgrade but don't have the default `SkipCertificateValidation` entry

Connects to #756


adamralph:
👍 good catch @WilliamBZA

PR #763, opened 2017-11-14
Turning the 'Contact Sales' button into a mailto link

### LINKED ISSUE #727: Make 'Contact Sales' option visible before end of trial
Currently in the license/registration dialog the Contact Sales option is disabled until the extended trial has expired. This option needs to remain visible throughout in case user wants to contact Sales even before the end of trial.

Connects to https://github.com/Particular/CustomerSuccess/issues/524

HEskandari:
Closes #727

### DIFF
--- a/src/ServiceInsight/Shell/LicenseRegistrationViewModel.cs
+++ b/src/ServiceInsight/Shell/LicenseRegistrationViewModel.cs
@@ -66,9 +66,7 @@
 
         public bool CanExtendTrial => HasTrialLicense && !licenseManager.CurrentLicense.IsExtendedTrial;
 
-        public bool CanBuyNow => CanContactSales;
-
-        public bool CanContactSales => HasTrialLicense && licenseManager.CurrentLicense.IsExtendedTrial;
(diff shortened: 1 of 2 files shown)


src/ServiceInsight/Shell/LicenseRegistrationViewModel.cs
```diff
@@ -123,7 +121,7 @@ public void Extend()
 
         public void ContactSales()
         {
-            network.Browse("http://particular.net/extend-your-trial-45");
+            network.Browse("mailto:contact@particular.net?subject=ServiceInsight");
```
WilliamBZA: What happens if no mail client is installed?

WilliamBZA: https://stackoverflow.com/questions/27176849/what-does-mailto-do-when-there-is-no-email-client

TL;DR: I don't think it's something we should concern ourselves with.

HEskandari: When balancing effort vs. value we decided to go with mailto instead of a webform on the website, so this is an acceptable risk (although Windows and Mac both come with mail clients preinstalled)

PR #769, opened 2018-01-22
Install Microsoft.SDK.Expression.Blend instead of relying on installed expression blend SDK

danielmarbach:
Tried to follow the build instructions explained in https://github.com/Particular/ServiceInsight#prerequisites but I could install Expression Blend SDK over the VS installer. I used 15.4.5.0 of VS.

But there seems to be a nuget package for the expression blend SDK. How about we use that one?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,5 @@ If using Visual Studio 2017, ensure that you have installed:
 
 - Workloads
   - .NET desktop development
-- Individual components
-  - Blend for Visual Studio SDK for .NET
 
 You can check whether the prerequisites have been installed by running the Visual Studio Installer and clicking "Modify".
-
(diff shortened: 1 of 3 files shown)


HEskandari:
I like the idea of depending on the nuget package instead of the installed component.

adamralph:
I do too. I'm just running a clean VM test...

adamralph:
Marking as WIP while the fixup needs squashing.

danielmarbach:
@adamralph thanks for the fixup. Squashed

HEskandari:
Consider updating the readme too (prerequisite section).

danielmarbach:
@HEskandari look closer ;) @adamralph added it

PR #771, opened 2018-01-26
Forward InstanceId to service control

danielmarbach:
In the multi-instance mode, the retry command in ServiceControl master needs to be sent either to the local instance or one of the remotes in case the remote did handle the error. 

Service Control newer versions of Service Control will return the instance id that can be passed back to ServiceControl. Based on the configured slave instances it then figures out to which instance it needs to send the message to. If the parameter is not present it always sends it to itself

### DIFF
--- a/src/ServiceInsight.Tests/Framework/Commands/RetryMessageCommandTests.cs
+++ b/src/ServiceInsight.Tests/Framework/Commands/RetryMessageCommandTests.cs
@@ -1,0 +1,44 @@
+﻿namespace ServiceInsight.Tests.Framework.Commands
+{
+    using Caliburn.Micro;
+    using NSubstitute;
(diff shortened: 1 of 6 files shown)


adamralph:
@danielmarbach what would be a good title for this issue in the context of our release notes? Is it effectively enabling SI to work with a new feature in SC vnext?

WilliamBZA:
It essentially completes the PR for https://github.com/Particular/ServiceInsight/pull/737

Which should be retitled to "Allow ServiceInsight to connect to scaled out instances of ServiceControl"

This particular PR is an optimization of that. It really has no value until that other PR is merged and the ServiceControl PRs are released.

danielmarbach:
@adamralph what @WilliamBZA said

PR #773, opened 2018-01-30
Allow ServiceInsight to connect to scaled out instances of ServiceControl

danielmarbach:
Replaces #737

Connects to #774

The PR contains the following changes:

- Removes SI opinionated way of paging and lets it just follow the predefined paging links that are returned by Service Control
- Removes the PageCount from the paging

### Paging

SC generates `Link` header which contains up to 4 links: next, prev, first and last. These are used, instead of generating the links in SI, to page through the data.

The main benefit is making SI not dependent on the actual paging URL format of SC. The SI only needs to know how to initiate the search.

The actual goal is to be able to insert an intermediary between SI and SC. That intermediary could aggregate the results from multiple instances of SC without SI even knowing about its existence.

### Page Count

As part of the TF work of extending SC with the multi instance mode for audits we discovered that the page count is not possible to determine upfront by doing a simple calculation based on the assumed page size. Since we went from a simplified paging approach in contrast to keep the offsets per SC instance each page can return up to `numberOfInstance * page_size` worth of data. Audits cannot be assumed to be uniformly distributed. Based on that we figured having a page size visible at all time makes no sense. The ability to page through the data is more important than knowing how many pages there are.

### DIFF
--- a/src/ServiceInsight.Tests/MessageListViewModelTests.cs
+++ b/src/ServiceInsight.Tests/MessageListViewModelTests.cs
@@ -51,7 +51,7 @@
         public void Should_load_the_messages_from_the_endpoint()
         {
             var endpoint = new Endpoint { Host = "localhost", Name = "Service" };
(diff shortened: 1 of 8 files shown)


adamralph:
Replaced by #775

Issue #774, opened 2018-01-30
Allow ServiceInsight to connect to scaled out instances of ServiceControl


adamralph:
### Paging

SC generates `Link` header which contains up to 4 links: next, prev, first and last. These are used, instead of generating the links in SI, to page through the data.

The main benefit is making SI not dependent on the actual paging URL format of SC. The SI only needs to know how to initiate the search.

The actual goal is to be able to insert an intermediary between SI and SC. That intermediary could aggregate the results from multiple instances of SC without SI even knowing about its existence.

### Page count

As part of the TF work of extending SC with the multi instance mode for audits we discovered that the page count is not possible to determine upfront by doing a simple calculation based on the assumed page size. Since we went from a simplified paging approach in contrast to keep the offsets per SC instance each page can return up to `numberOfInstance * page_size` worth of data. Audits cannot be assumed to be uniformly distributed. Based on that we figured having a page size visible at all time makes no sense. The ability to page through the data is more important than knowing how many pages there are.

### Known endpoints

Previously ServiceInsight was using `/endpoints` which is essentially a mixture of multiple concerns. For example, the data structure returned there contains information like whether monitoring of the endpoint is on which is not relevant for SI. With the known endpoints API there are two distinct endpoints now. One for SI which contains only the SI relevant information and one for SP.

## Plan of attack

- [x] Switch to link based paging - #775
- [x] Remove page count  - #775
- [x] Supply instance ID for retries  - #771
- ~[ ] Supply instance ID for body fetching~ No longer required since SC will return the instance id in the provided body URL
- [x] Use endpoints/known when service is discovered - #778 
- [x] Async UI for better responsiveness - #781

adamralph:
> Async UI for better responsiveness - #781

PR merged and PoA item checked.

adamralph:
> Use endpoints/known when service is discovered - #778

PR merged and last PoA item checked. Closing.

PR #775, opened 2018-01-30
Paging and page count for scaled out ServiceControl

adamralph:
Connects to #774

### DIFF
--- a/src/ServiceInsight/Search/SearchBarViewModel.cs
+++ b/src/ServiceInsight/Search/SearchBarViewModel.cs
@@ -32,7 +32,6 @@
         {
             this.commandLineArgParser = commandLineArgParser;
             this.settingProvider = settingProvider;
-            PageSize = 50; //NOTE: Do we need to change this?
 
             SearchCommand = Command.Create(this, Search, vm => vm.CanSearch);
             CancelSearchCommand = Command.Create(this, CancelSearch, vm => vm.CanCancelSearch);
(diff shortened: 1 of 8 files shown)


adamralph:
@danielmarbach any objection to merging?

WilliamBZA:
Shouldn't this be squashed before merging @adamralph?

adamralph:
@WilliamBZA no, there are no fixups here. Each commit is self contained.

adamralph:
I just noticed that ne of the commit messages has a weird typo:

> Refactor: Introduce MessageList/MessageListViewModel.RefreshMessages(string, bool)

It should be:

> Refactor: Introduce MessageListViewModel.RefreshMessages(string, bool)

I can fix that before we merge if preferred.

src/ServiceInsight/Search/SearchBarViewModel.cs
```diff
@@ -52,22 +51,22 @@ protected override void OnActivate()
...
         public void GoToNextPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, CurrentPage + 1, SearchQuery);
+            Parent.RefreshMessages(NextLink);
         }
 
         public void GoToLastPage()
         {
-            Parent.RefreshMessages(SelectedEndpoint, PageCount, SearchQuery);
+            Parent.RefreshMessages(LastLink);
```
danielmarbach: I must say I preferred the previous name "NavigateTo" but that is just my taste

danielmarbach:
@adamralph go ahead, I approved

src/ServiceInsight/Search/SearchBarViewModel.cs
adamralph: That's also refactoring we could consider. It doesn't have to be part of this PR.

PR #777, opened 2018-02-01
Upgrade caliburn.micro to 3.2.0

danielmarbach:

### DIFF
--- a/src/ServiceInsight/ServiceInsight.csproj
+++ b/src/ServiceInsight/ServiceInsight.csproj
@@ -65,13 +65,14 @@
     <Reference Include="Autofac.Configuration">
       <HintPath>..\packages\Autofac.Configuration.3.3.0\lib\net40\Autofac.Configuration.dll</HintPath>
     </Reference>
-    <Reference Include="Caliburn.Micro, Version=2.0.2.0, Culture=neutral, PublicKeyToken=8e5891231f2ed21f, processorArchitecture=MSIL">
-      <HintPath>..\packages\Caliburn.Micro.Core.2.0.2\lib\net45\Caliburn.Micro.dll</HintPath>
-      <Private>True</Private>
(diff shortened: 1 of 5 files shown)


src/ServiceInsight/ServiceInsight.csproj
```diff
@@ -239,7 +240,7 @@
     <Reference Include="System.Transactions" />
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Windows.Interactivity, Version=4.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\packages\Microsoft.SDK.Expression.Blend.1.0.0\lib\net45\System.Windows.Interactivity.dll</HintPath>
+      <HintPath>..\packages\Caliburn.Micro.3.2.0\lib\net45\System.Windows.Interactivity.dll</HintPath>
```
adamralph: Does this mean we can remove the installation of `Microsoft.SDK.Expression.Blend` that was added in #769?

danielmarbach:
I did smoke test the tabs, the properties window and the paging yes

src/ServiceInsight/ServiceInsight.csproj
danielmarbach: The expression blend SDK was removed automatically by nuget. I'm guessing yes.

danielmarbach:
@HEskandari it wouldn't hurt. But I wonder why this has to be released just and why we can't accumulate a few more things on master and then release it together

HEskandari:
For occasions we have more things coming through, we can delay the release a bit, but otherwise, we have 'release as soon as possible' rule of thumb.

WilliamBZA:
Features or bugs should be released once ready. Things like this offer no tangible value to users so should not fall within the same rules as features or bugs.

PR #778, opened 2018-02-01
Prefer KnownEndpointsUrl, to EndpointsUrl, to local default

boblangley:

We have introduced a distinct API call to return the known endpoints from ServiceControl which is the, now, more correct call for ServiceInsight to make.

The change is intended to be backwards compatible, falling back to the "endpoints" const string when no other service is reported by ServiceControl for SI to use.

### DIFF
--- a/src/ServiceInsight/Models/Endpoint.cs
+++ b/src/ServiceInsight/Models/Endpoint.cs
@@ -2,23 +2,31 @@
 {
     public class Endpoint
     {
-        string host;
-
-        public string Name { get; set; }
+        public string Name
+        {
+            get => EndpointDetails.Name;
+            set => EndpointDetails.Name = value;
+        }
 
         public string Host
         {
-            get { return host ?? HostDisplayName; }
-            set { host = value; }
+            get => EndpointDetails.Host ?? HostDisplayName;
+            set => EndpointDetails.Host = value;
         }
 
-        public string HostId { get; set; }
+        public string HostId
+        {
+            get => EndpointDetails.HostId;
+            set => EndpointDetails.HostId = value;
+        }
 
         public string HostDisplayName { get; set; }
 
-        public string Address => string.Format("{0}{1}", Name, AtMachine());
+        public string Address => $"{Name}{AtMachine()}";
 
-        string AtMachine() => string.IsNullOrEmpty(Host) ? string.Empty : string.Format("@{0}", Host);
+        public EndpointDetails EndpointDetails { get; set; } = new EndpointDetails();
+
+        string AtMachine() => string.IsNullOrEmpty(Host) ? string.Empty : $"@{Host}";
 
         protected bool Equals(Endpoint other) => other != null && string.Equals(Address, other.Address);
 
@@ -49,5 +57,14 @@
         public static bool operator !=(Endpoint left, Endpoint right) => !Equals(left, right);
 
         public override string ToString() => Address;
+    }
+
+    public class EndpointDetails
+    {
+        public string Name { get; set; }
+
+        public string HostId { get; set; }
+
+        public string Host { get; set; }
     }
 }
--- a/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
+++ b/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
@@ -148,8 +148,8 @@ namespace ServiceInsight.ServiceControl
             endpointsUrl = endpointsUrl.Replace(connection.Url, string.Empty);
 
             var request = new RestRequestWithCache(endpointsUrl, RestRequestWithCache.CacheStyle.IfNotModified);
-            var messages = await GetModel<List<Endpoint>>(request).ConfigureAwait(false);
-            return messages ?? new List<Endpoint>();
+            var endpoints = await GetModel<List<Endpoint>>(request).ConfigureAwait(false);
+            return endpoints ?? new List<Endpoint>();
         }
 
         public async Task<IEnumerable<KeyValuePair<string, string>>> GetMessageData(SagaMessage message)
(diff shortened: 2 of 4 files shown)


src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -134,7 +141,16 @@ public IEnumerable<StoredMessage> GetConversationById(string conversationId)
...
         public IEnumerable<Endpoint> GetEndpoints()
         {
-            var request = new RestRequestWithCache(EndpointsEndpoint, RestRequestWithCache.CacheStyle.IfNotModified);
+            var rootUrls = GetRootUrls();
+
+            var endpointsUrl = rootUrls.KnownEndpointsUrl ?? rootUrls.EndpointsUrl ?? DefaultEndpointsEndpoint;
+
+            if (endpointsUrl.Contains(connection.Url))
+            {
+                endpointsUrl = endpointsUrl.Replace(connection.Url, string.Empty);
```
danielmarbach: could we just use replace without the contains?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -81,6 +81,13 @@ public string GetVersion()
             return header == null ? null : header.Value.ToString();
         }
 
+        ServiceControlRootUrls GetRootUrls()
```
danielmarbach: @Particular/serviceinsight-maintainers this API could be used to replace all the hardcoded URIs in the future. We figured that is something that can be done at a later stage as a refactoring

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -134,7 +141,16 @@ public IEnumerable<StoredMessage> GetConversationById(string conversationId)
 
         public IEnumerable<Endpoint> GetEndpoints()
         {
-            var request = new RestRequestWithCache(EndpointsEndpoint, RestRequestWithCache.CacheStyle.IfNotModified);
+            var rootUrls = GetRootUrls();
+
+            var endpointsUrl = rootUrls.KnownEndpointsUrl ?? rootUrls.EndpointsUrl ?? DefaultEndpointsEndpoint;
```
adamralph: As discussed, we will put the Elvis operator into the `rootUrls` calls.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
adamralph: As discussed: yes.

adamralph:
Marking as WIP because this needs some more smoke tests.

adamralph:
Pushed a fix up with the changes we discussed. Still a WIP while we smoke test and before we squash.

SeanFeldman:
@boblangley and I have fixed an issue when SC was returning KnownEndpointView rather than EndpointView (older API). Smoked tested with this PR and a fallback (again `master`). Both render endpoints as expected. @danielmarbach @adamralph please review.

danielmarbach:
@SeanFeldman @boblangley heads up I rebased this on top of the async changes introduced in #781  You need to reset your branches locally

WilliamBZA:
Is this still WIP?

src/ServiceInsight/Models/Endpoint.cs
```diff
@@ -2,23 +2,31 @@
 {
     public class Endpoint
     {
-        string host;
-
-        public string Name { get; set; }
+        public string Name
+        {
+            get => EndpointDetails.Name;
```
HEskandari: This uses expression assignment instead of old get/set style. Not that I don't like it but for consistency, I think we need to keep using the older/established coding style. Same goes for below.

src/ServiceInsight/Models/Endpoint.cs
```diff
@@ -2,23 +2,31 @@
...
+        public string HostId
+        {
+            get => EndpointDetails.HostId;
+            set => EndpointDetails.HostId = value;
+        }
 
         public string HostDisplayName { get; set; }
 
-        public string Address => string.Format("{0}{1}", Name, AtMachine());
+        public string Address => $"{Name}{AtMachine()}";
```
HEskandari: This does not seem to be related to the task at hand. Suggesting to keep as separate refactoring commit/PR.

src/ServiceInsight/Models/Endpoint.cs
```diff
@@ -2,23 +2,31 @@
...
+
+        public EndpointDetails EndpointDetails { get; set; } = new EndpointDetails();
 
-        string AtMachine() => string.IsNullOrEmpty(Host) ? string.Empty : string.Format("@{0}", Host);
+        string AtMachine() => string.IsNullOrEmpty(Host) ? string.Empty : $"@{Host}";
```
HEskandari: This does not seem to be related to the task at hand. Suggesting to keep as separate refactoring commit/PR.

danielmarbach:
Not from my side. I tested this locally but I think @adamralph marked it as wip for more testing

src/ServiceInsight/Models/Endpoint.cs
danielmarbach: @HEskandari why? The class was touched in this PR. While not ideal I don't see why we can't leave it as is

src/ServiceInsight/Models/Endpoint.cs
adamralph: Personaly I'd rather see us move everything gradually toward expression bodied members. We should probably be consistent within a single file, though, and I think we are here.

adamralph: Discussed this with @danielmarbach and we think this is better, and it is consistent with other existing members in this file.

src/ServiceInsight/Models/Endpoint.cs
adamralph: Discussed with @danielmarbach and we agree that this is minor enough that we can leave as-is this time.

src/ServiceInsight/Models/Endpoint.cs
adamralph: Discussed with @danielmarbach and we agree that this is minor enough that we can leave as-is this time.

PR #780, opened 2018-02-02
Set current page to one instead of zero

danielmarbach:
Fixes weird behavior introduced by refactoring according to this comment 

https://github.com/Particular/ServiceInsight/pull/737#discussion_r164502253

### Before

150 msgs

Start

image

click next

image

indicates page 2

go back

image

indicates page 1

### After

Right start with page 1

image

Empty result set is still page 0

image

Connects to #774.


adamralph:
This was introduced during the work for #774 so I connected it to that issue.

PR #781, opened 2018-02-02
Async Handling of ServiceControl communication

### LINKED ISSUE #779: Freezing when querying data from ServiceControl
The UI freezes when querying data from ServiceControl. If the instance of ServiceControl is slow then UI freezes for a frustratingly long time.

The query methods should be async with appropriate loading animations.

danielmarbach:
Fixes #779 

- RestSharp upgrade is required. The version used on master throws webexceptions in the async mode

I've smoke tested it with one and multiple ServiceControl instances. The overall behavior is much smoother now even when the master SC is timing out. As a plus the loading screens are now finally visible again because the UI has time to actually render those :)

The video shows the behavior with a multi instance master where the slave is not available and timing out

[ServiceInsight Async](https://youtu.be/g7m2BBXkag0)

The video below shows the behavior with multi-instance when the slaves are available

[ServiceInsight Async](https://youtu.be/WFfeXsyVYsM)

You'll see once how the sequence diagram collapses. I have that on my machine even with the current released version.

### DIFF
--- a/src/ServiceInsight.Tests/SearchViewModelTests.cs
+++ b/src/ServiceInsight.Tests/SearchViewModelTests.cs
@@ -1,6 +1,7 @@
 ﻿namespace ServiceInsight.Tests
 {
     using System;
+    using System.Windows.Threading;
     using NSubstitute;
     using NUnit.Framework;
     using ServiceInsight.Explorer.EndpointExplorer;
@@ -21,11 +22,15 @@
             var argParser = Substitute.For<CommandLineArgParser>();
             var settingProvider = Substitute.For<ISettingsProvider>();
 
+            // ReSharper disable once UnusedVariable
+            // required for async command
+            var dispatcher = Dispatcher.CurrentDispatcher;
             viewModel = new SearchBarViewModel(argParser, settingProvider);
         }
 
         public void Dispose()
         {
+            Dispatcher.ExitAllFrames();
             viewModel.Dispose();
         }
 
--- a/src/ServiceInsight/Framework/Command.cs
+++ b/src/ServiceInsight/Framework/Command.cs
@@ -3,22 +3,35 @@
     using System;
     using System.ComponentModel;
     using System.Reactive.Linq;
+    using System.Threading.Tasks;
     using System.Windows.Input;
     using Commands;
     using ExtensionMethods;
 
     public static class Command
     {
+        public static ICommand CreateAsync(Func<Task> executeAction, IObservable<bool> canExecuteObservable) =>
+            new AsyncObservableCommand(canExecuteObservable, _ => executeAction());
+
         public static ICommand Create(Action executeAction, IObservable<bool> canExecuteObservable) =>
             new ObservableCommand(canExecuteObservable, _ => executeAction());
 
         public static ICommand Create<TPropertyChanged>(TPropertyChanged propertyChanged, Action executeAction, Func<TPropertyChanged, bool> canExecuteSelector) where TPropertyChanged : INotifyPropertyChanged =>
             Create(executeAction, propertyChanged.ChangedProperty().Select(_ => canExecuteSelector(propertyChanged)));
 
+        public static ICommand CreateAsync<TPropertyChanged>(TPropertyChanged propertyChanged, Func<Task> executeAction, Func<TPropertyChanged, bool> canExecuteSelector) where TPropertyChanged : INotifyPropertyChanged =>
+            CreateAsync(executeAction, propertyChanged.ChangedProperty().Select(_ => canExecuteSelector(propertyChanged)));
+
         public static ICommand Create(Action executeAction) =>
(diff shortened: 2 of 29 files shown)


src/ServiceInsight.Tests/SearchViewModelTests.cs
```diff
@@ -21,11 +22,15 @@ public SearchViewModelTests()
             var argParser = Substitute.For<CommandLineArgParser>();
             var settingProvider = Substitute.For<ISettingsProvider>();
 
+            // ReSharper disable once UnusedVariable
+            // required for async command
+            var dispatcher = Dispatcher.CurrentDispatcher;
```
adamralph: From https://msdn.microsoft.com/en-us/library/system.windows.threading.dispatcher.currentdispatcher(v=vs.110).aspx

> Gets the Dispatcher for the thread currently executing and creates a new Dispatcher if one is not already associated with the thread.

@danielmarbach will add more context.

src/ServiceInsight/Framework/Command.cs
```diff
@@ -3,22 +3,35 @@
...
+    using System.Threading.Tasks;
     using System.Windows.Input;
     using Commands;
     using ExtensionMethods;
 
     public static class Command
     {
+        public static ICommand CreateAsync(Func<Task> executeAction, IObservable<bool> canExecuteObservable) =>
+            new AsyncObservableCommand(canExecuteObservable, _ => executeAction());
+
```
adamralph: Nitpick: this could be below the sync version as with the others.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -60,9 +61,9 @@ static DefaultServiceControl()
             cache = new MemoryCache("ServiceControlReponses", new NameValueCollection(1) { { "cacheMemoryLimitMegabytes", settings.CacheSize.ToString() } });
         }
 
-        public bool IsAlive()
+        public async Task<bool> IsAlive()
         {
-            var version = GetVersion();
+            var version = await GetVersion().ConfigureAwait(false);
```
adamralph: We only call `ConfigureAwait(false)` here, because this code has no knowledge of the UI and will therefore never use the UI thread.

HEskandari:
Here is the issue I talked about: see how three requests are logged (which is kind of okay on its own) but the response DEBUG logs with no correlation to the request could cause confusion.

image

src/ServiceInsight.Tests/SearchViewModelTests.cs
danielmarbach: The SearchViewModel uses a ReactiveCommand that triggers the execute changed over Rx. Due to the context switch with async the RxObservable needs to be executed back on the UI thread. This is a strick here because accessing the CurrentDispatcher creates one that is associated with the current thread if there is not already one available thus allowing to execute the execute changed on a dispatcher.

adamralph:
(I've marked this as WIP while we discuss the logging issue.)

How about we append the request URI to the end of every "Response Header" log message? Then they can be correlated.

danielmarbach:
Rebased on top of master.

@Particular/serviceinsight-maintainers while it is possible to correlate things, for example by augmenting the log events with properties, using the URI as a grouping mechanism or using an interlocked increment and then pass that on we can't really group things in the observable stream. We could by tradding off memory resort the observable collection every time when a new LogEvent is added by the correlation id. But the observable collection is not sortable so we'd have to recreate it all the time which creates a lot of garbage. Another tradeoff would be to move the LogRequest Method in the DefaultServiceControl implementation below the `restClient.ExecuteTaskAsync` method. Obviously, this has the drawback that you don't see anything pending while a rest request is pending.

adamralph:
What if we logged all the response headers in the same log message as the OK? That message already has the request URI in it. It would be a long line, which would wrap, but does that matter? E.g.

> 2018-02-09 20:01:48.461 +02:00 [Debug] HTTP Status OK (http://localhost:33333/api/sagas/3058ae11-c30a-4fd9-9fcd-a87a00839167) { "Transfer-Encoding" : "chunked", "Content-Encoding" : "", "Vary" : "Accept" ... }

adamralph:
@HEskandari and I discussed this and we think putting the headers into the same message as the response OK will be fine.

@WilliamBZA @danielmarbach any objections?

danielmarbach:
Pragmatic and good for me. I see no drawbacks. The UX experience shoukd not
suffer

Am 12.02.2018 16:12 schrieb "Adam Ralph" <notifications@github.com>:

SeanFeldman:
Updated PR to output what @adamralph suggested:

image

@adamralph @danielmarbach @HEskandari would this work for you?

HEskandari:
@SeanFeldman works for me. I give the PR my blessing.

danielmarbach:
LGTM

PR #788, opened 2018-02-22
fix endpoint retrieval to cater for differing public/private host and port

### LINKED ISSUE #787: Endpoint Explorer using invalid url when retrieving endpoints (SI 1.10.0)
In version 1.10.0 the Endpoint Explorer is not populated with data.

The logs shows that an invalid url is used when retrieving the data from SC:

billede

adamralph:
fixes #787

### DIFF
--- a/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
+++ b/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
@@ -145,7 +145,11 @@ namespace ServiceInsight.ServiceControl
             var rootUrls = await GetRootUrls().ConfigureAwait(false);
 
             var endpointsUrl = rootUrls?.KnownEndpointsUrl ?? rootUrls?.EndpointsUrl ?? DefaultEndpointsEndpoint;
-            endpointsUrl = endpointsUrl.Replace(connection.Url, string.Empty);
+            if (Uri.TryCreate(endpointsUrl, UriKind.RelativeOrAbsolute, out var endpointsUri) &&
+                Uri.TryCreate(connection.Url, UriKind.RelativeOrAbsolute, out var connectionUri))
+            {
+                endpointsUrl = endpointsUri.PathAndQuery.Replace(connectionUri.PathAndQuery, string.Empty);
+            }
 
             var request = new RestRequestWithCache(endpointsUrl, RestRequestWithCache.CacheStyle.IfNotModified);
             var endpoints = await GetModel<List<Endpoint>>(request).ConfigureAwait(false);


src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -145,7 +145,11 @@ public Task<PagedResult<StoredMessage>> GetAuditMessages(string link)
             var rootUrls = await GetRootUrls().ConfigureAwait(false);
 
             var endpointsUrl = rootUrls?.KnownEndpointsUrl ?? rootUrls?.EndpointsUrl ?? DefaultEndpointsEndpoint;
-            endpointsUrl = endpointsUrl.Replace(connection.Url, string.Empty);
+            if (Uri.TryCreate(endpointsUrl, UriKind.RelativeOrAbsolute, out var endpointsUri) &&
```
danielmarbach: So if we can't create the URIs we will use whatever that was provided. Can we envision any scenarios where that would fail?

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -145,7 +145,11 @@ public Task<PagedResult<StoredMessage>> GetAuditMessages(string link)
...
+                Uri.TryCreate(connection.Url, UriKind.RelativeOrAbsolute, out var connectionUri))
+            {
+                endpointsUrl = endpointsUri.PathAndQuery.Replace(connectionUri.PathAndQuery, string.Empty);
```
danielmarbach: So if we get

```
/api/endpoints/
```

and the connection URI is

```
http://localhost:33333/api
```

the replace will lead to

```
http://localhost:33333/api/endpoints/
```

if

```
http://localhost:33333/weird/api/weird/api
```

is provided the PathAndQuer is `/weird/api/weird/api` and endpoints URI PathAndQuery would be `/api/endpoints`  which then would lead to an incorrect URI again with the replace right?

adamralph: @danielmarbach you are correct that if a proxy changes the path somehow, e.g. inserting path components before the path known by SC, then this won't work.

This fix only works when the use of a proxy means that SI and SC are only using a different hostname and port. Specifically, it will fix the problem reported in https://github.com/Particular/ServiceInsight/issues/787.

If the more exotic case arises where a proxy is changing the path, we will have to consider another fix (such as ignoring absolute URI's from SC completely and falling back to our hard coded default), but I suggest we try this patch first since it's unlikely anyone will be affected by the more exotic edge case.

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
adamralph: In that case, both the SI connection URL and the SC URL are garbage, and things should rightly blow up.

adamralph: Note that there is no safeguard against that in 1.9.1, 1.10.0 or after this change. The queries will fail in all cases and the log output is all we have to go on.

PR #797, opened 2018-06-05
Truncating lists only for Saga data related operations

### LINKED ISSUE #795: ServiceInsight ignores datasets larger than 200 items
The issue boils down to this line of code:

https://github.com/Particular/ServiceInsight/blob/1cc9820086de31ec752e7d7d3b5d03dca60f9ec2/src/ServiceInsight/Framework/MessageDecoders/JsonMessageDeserializer.cs#L140

This deserializer is used across the board and is not just Saga Data that uses it (which I think was the assumption). What happens is that if Endpoint List (or any other ServiceControl endpoint for that matter) returns a large list (greater than 200), it will get cropped.

HEskandari:
Resolves #795

### DIFF
--- a/src/ServiceInsight/ExtensionMethods/RestClientExtensions.cs
+++ b/src/ServiceInsight/ExtensionMethods/RestClientExtensions.cs
@@ -1,0 +1,23 @@
+﻿namespace ServiceInsight.ExtensionMethods
+{
+    using RestSharp;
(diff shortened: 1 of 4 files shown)


WilliamBZA:
Looks good, this can be squashed now

HEskandari:
@WilliamBZA  This is _now_ squashed.

PR #799, opened 2018-06-06
Capturing SSL cert validation errors and showing proper error

### LINKED ISSUE #793: Error notification for certificate validation failure
I just upgraded a machine from ServiceInsight 1.8.0 to 1.10.2 and suddenly I couldn't connect to an endpoint anymore. I was getting an error saying unable to connect to ServiceControl. 

I dug through the code and narrowed it down to the certificate validation flag added in the app.config. When I changed SkipCertificateValidation to true, it started working, but it wasn't obvious what the problem was. It would have been nice if ServiceInsight told me there was a certificate issue.

HEskandari:
Resolves #793

### DIFF
--- a/src/ServiceInsight/Shell/ServiceControlConnectionViewModel.cs
+++ b/src/ServiceInsight/Shell/ServiceControlConnectionViewModel.cs
@@ -13,11 +13,19 @@
 
     public class ServiceControlConnectionViewModel : Screen
     {
-        public const string ConnectingToServiceControl = "Connecting to ServiceControl...";
+        const string ConnectingToServiceControl = "Connecting to ServiceControl...";
+        const string ConnectionErrorMessage = "There was an error connecting to ServiceControl. Either the address is not valid or the service is down.";
+        const string CertValidationErrorMessage = "There was an error connecting to ServiceControl. SSL certificate is not valid.";
+        static bool certValidationFailed;
 
         ISettingsProvider settingsProvider;
         ProfilerSettings appSettings;
         IContainer container;
+
+        static ServiceControlConnectionViewModel()
+        {
+            DefaultServiceControl.CertificateValidationFailed = () => { certValidationFailed = true; };
+        }
 
         public ServiceControlConnectionViewModel(
             ISettingsProvider settingsProvider,
@@ -31,7 +39,9 @@
 
         public string ServiceUrl { get; set; }
 
-        public bool IsAddressValid { get; set; }
+        public bool ShowError { get; set; }
+
+        public string ErrorMessage { get; set; }
 
         public bool WorkInProgress { get; private set; }
 
@@ -39,7 +49,7 @@
         {
             base.OnActivate();
 
-            IsAddressValid = true;
+            ShowError = false;
             ServiceUrl = appSettings.LastUsedServiceControl;
             RecentEntries = GetRecentServiceEntries();
         }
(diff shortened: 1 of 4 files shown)


src/ServiceInsight/Shell/ServiceControlConnectionViewModel.cs
```diff
@@ -13,12 +14,26 @@
...
+        const string CertValidationErrorMessage = "There was an error connecting to ServiceControl. SSL certificate is not valid.";
+        static bool certValidationFailed;
 
         ISettingsProvider settingsProvider;
         ProfilerSettings appSettings;
         IContainer container;
 
+        static ServiceControlConnectionViewModel()
+        {
+            ServicePointManager.ServerCertificateValidationCallback = (sender, certificate, chain, errors) => OnCertificateValidationFailed();
```
Scooletz: This is a static delegate, but we assign for this `ServiceControlConnectionViewModel` only. This doesn't look right to me. I followed `IsValidUrl` and as far as I understand, eventually, it calls `DefaultServiceControl`, which underneath issues a rest request against the specified `url`. I know that passing the delegate down the drain, till it meets the `IRestClient` implementation might be hard, but we'd ensure that we don't loose anything (or override something for future versions).

HEskandari: Problem is that the callback is static. If you look more, DefaultServiceControl class has also subscribed to the same delegate. Now we have some options:

- Move the delegate to the ViewModel class? It doesn't sound right as it is not _just_ a presentation concern.

- Have two delegates attached to the callback? This is what we have. The callback uses the _last_ returned result, so since we don't control which one is the last one, I had to use the exact same code (e.g. return the same flag) to ensure it follows the same logic.

- Have another static callback in DefaultServiceControl which gets called as a part of ServerCertificateValidationCallback. That helps with duplicate `return` logic, but still not great.

I'll update the PR with the third option, see what you think.

HEskandari:
@Scooletz Please review the update.

Scooletz:
I think this is a move into a better direction. Still, I'd think about using the `HttpRequest.ServerCertificateValidationCallback` if possible https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest.servercertificatevalidationcallback(v=vs.110).aspx
To me it looks like a question to a maintainer, how much should be done in this PR. From the PR perspective it looks ok. From the repo, maybe it should be moved a bit further to the http request (if possible).
I'm approving it as it's a good version we can live with to address the fix.

src/ServiceInsight/Shell/ServiceControlConnectionViewModel.cs
```diff
@@ -59,14 +69,20 @@ public virtual void Close()
 
         public virtual async Task Accept()
         {
+            certValidationFailed = false;
             StartWorkInProgress();
             ServiceUrl = ServiceUrl.Trim();
-            IsAddressValid = await IsValidUrl(ServiceUrl);
-            if (IsAddressValid)
+            ShowError = await IsValidUrl(ServiceUrl) == false;
+            if (!ShowError)
```
WilliamBZA: I think it would be more readable if the showError calculation was in the if statement instead of in the await.

```CSharp
var isUrlValid = await IsValidUrl(ServiceUrl);
ShowError = !isUrlValid;
```

HEskandari: refactored.

HEskandari:
@WilliamBZA Code refactored and tests compile and all pass now.

PR #800, opened 2018-06-06
Making payload and exception views resizable and storing the size

### LINKED ISSUE #714: Saga payload view should be resizable
Saga Payload view and Exception view should be resizable so that the user can resize them to fit their screen size and content.

HEskandari:
Resolves #714


Scooletz:
Too much WPF for me. I cannot confirm that it solves the issue nor that is implemented according standards.

PR #801, opened 2018-06-06
Changing the message selection when FlowDiagram popup is opened

### LINKED ISSUE #471: Clicking the link that opens the popup does not select the message
2015-06-22 12_20_31-serviceinsight for nservicebus

HEskandari:
Resolves #471


Scooletz:
Too much WPF for me. I cannot confirm that it solves the issue nor that is implemented according standards.

PR #802, opened 2018-06-06
Custom control to render white space correctly in Saga Update control

### LINKED ISSUE #668: Strings are not represented correctly in SagaView
There are various issues related to displaying values, since there's no type information and everything is a string.

- Empty strings are left empty on right hand side of the equal signs ("="). The lack of value should be clearly communicated.
- Trailing white space is not represented at all.
- You can't tell if empty space is an empty string or a series of spaces (or potentially tabs?)
- _Leading_ whitespace, in a way, is shown, by the offset of the RHS value but hard to tell how many white spaces there are.

HEskandari:
Resolves #668


Scooletz:
Too much WPF for me. I cannot confirm that it solves the issue nor that is implemented according to standards in the repo.

seanfarmar (approved):
LGTM Reviewed with @HEskandari

PR #803, opened 2018-06-07
Showing all validation errors when importing a license

### LINKED ISSUE #502: ServiceInsight does not provide feedback to the user when a license is validated
The license error message should contain information about which condition has not been met to help the customer in identifying the license issue. The error messages should be displayed in ServiceInsight when importing the license.

HEskandari:
Resolves #502

### DIFF
--- a/src/ServiceInsight/Framework/Licensing/AppLicenseManager.cs
+++ b/src/ServiceInsight/Framework/Licensing/AppLicenseManager.cs
@@ -15,33 +15,22 @@
 
         public bool TryInstallLicense(string licenseText)
         {
-            try
+            ValidationResult = LicenseDialogSource.Validate(licenseText);
+            if (ValidationResult.License != null)
             {
-                if (!LicenseVerifier.TryVerify(licenseText, out _))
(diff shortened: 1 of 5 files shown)


src/ServiceInsight/Framework/Licensing/AppLicenseManager.cs
```diff
@@ -56,4 +45,20 @@ public int GetRemainingTrialDays()
 
         public bool IsLicenseExpired() => LicenseExpirationChecker.HasLicenseExpired(CurrentLicense);
     }
+
+    internal class InMemoryLicenseSource : LicenseSource
```
WilliamBZA: Where is this used? I can't see where this is added to the result returned by `GetStandardLicenseSources`

HEskandari: Removed. I had refactored to use `LicenseDialogSource` instead.

HEskandari:
@WilliamBZA anything left here? should I squash?

PR #828, opened 2019-05-23
Supports passing machine name to app invocation

### LINKED ISSUE #827: Application Invocation URIs only support fully qualified domain names or localhost
## Who's affected

Customers using application invocations links like `si://anyknownserver:33333/api` and ServiceControl APIs being hosted under URIs like `http://anyknownserver:33333/api`

## Symptoms

Because we use isolated storage for storing profiles it is difficult to use register multiple ServiceControl URIs in the profile settings. As a workaround, we suggest using 

https://docs.particular.net/serviceinsight/application-invocation

In theory links like `si://anyknownserver:3333/api` for example should open ServiceInsight and connect to `http://anyknownserver:3333/api`

Unfortunately, there is an issue with this approach:

The regular expression used to validate the URI passed over the command line arguments enforces the usage of TLD, so links like `http://anyknownserver:33333/api` do not satisfy the regex but only fully qualified domain names would

Furthermore when the regex doesn't match SI silently falls back to ignore the URI and switches to use the recently used connection URI.

HEskandari:
Resolves #827

### DIFF
--- a/src/ServiceInsight/Models/CommandLineOptions.cs
+++ b/src/ServiceInsight/Models/CommandLineOptions.cs
@@ -1,13 +1,22 @@
 ﻿namespace ServiceInsight.Models
 {
     using System;
-    using System.Text.RegularExpressions;
+    using System.Collections.Generic;
+    using System.Linq;
     using RestSharp.Extensions.MonoHttp;
 
     public class CommandLineOptions
     {
         public const string ApplicationScheme = "si://";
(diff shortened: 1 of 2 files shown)


src/ServiceInsight/Models/CommandLineOptions.cs
```diff
@@ -9,6 +12,15 @@ public class CommandLineOptions
         public const string ApplicationScheme = "si://";
         public const string UriRegexPattern = @"^(http|https)\://([a-zA-Z0-9\.\-]+(\:[a-zA-Z0-9\.&amp;%\$\-]+)*@)*((25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9])\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[0-9])|localhost|([a-zA-Z0-9\-]+\.)*[a-zA-Z0-9\-]+\.(com|edu|gov|int|mil|net|org|biz|arpa|info|name|pro|aero|coop|museum|[a-zA-Z]{2}))(\:[0-9]+)*(/($|[a-zA-Z0-9\.\,\?\'\\\+&amp;%\$#\=~_\-]+))*$";
```
mikeminutillo: Is this still used?

HEskandari: Good point.

HEskandari:
@mikeminutillo updated.

PR #831, opened 2019-05-29
support multi-saga starts in flow diagram

### LINKED ISSUE #230: Visualization of a message starting multiple Sagas in the Flow Diagram
In the `Message Flow` diagram when the message starts more than one saga, the message node should correctly visualize all the Sagas. 

For example, the message starting two sagas should look like this:

Screen Shot 2019-05-29 at 12 58 27 pm

HEskandari:
Resolves #230

### DIFF
--- a/src/ServiceInsight/MessageFlow/MessageFlowView.xaml
+++ b/src/ServiceInsight/MessageFlow/MessageFlowView.xaml
@@ -364,32 +364,68 @@
                                                         <local:MessageActionPopup />
                                                     </local:MessagePopupControl>
 
-                                                    <Grid Grid.Row="2"
-                                                          Margin="0 -5"
-                                                          Visibility="{Binding HasSaga,
-                                                                               Converter={StaticResource BoolToVisibilityConverter}}"
-                                                          cal:Message.Attach="[Event MouseLeftButtonUp]=[Action ShowSagaWindow]">
-                                                        <Grid.ColumnDefinitions>
-                                                            <ColumnDefinition Width="20" />
-                                                            <ColumnDefinition Width="*" />
-                                                        </Grid.ColumnDefinitions>
-                                                        <ContentControl x:Name="SagaStatusIcon"
-                                                                        HorizontalAlignment="Center"
-                                                                        VerticalAlignment="Center"
-                                                                        Template="{StaticResource SagaInitiatedIcon}" />
-                                                        <Border Grid.Column="1"
-                                                                Margin="2 0 0 0"
-                                                                Background="#333333">
-                                                            <TextBlock VerticalAlignment="Center"
-                                                                       FontSize="11px"
-                                                                       FontWeight="Bold"
-                                                                       Foreground="{StaticResource MessageBackground}"
-                                                                       Padding="3 0"
-                                                                       Text="{Binding SagaType}"
-                                                                       TextTrimming="CharacterEllipsis"
-                                                                       TextWrapping="NoWrap" />
-                                                        </Border>
-                                                    </Grid>
+                                                    <ItemsControl ItemsSource="{Binding SagaInvocations}"
+                                                                 Grid.Row="2"
+                                                                 Margin="0 -5 2 0"
+                                                                 Visibility="{Binding HasSaga, Converter={StaticResource BoolToVisibilityConverter}}"
+                                                                 cal:Message.Attach="[Event MouseLeftButtonUp]=[Action ShowSagaWindow]">
+                                                        <ItemsControl.ItemsPanel>
+                                                            <ItemsPanelTemplate>
+                                                                <UniformGrid Columns="1" IsItemsHost="True" />
+                                                            </ItemsPanelTemplate>
+                                                        </ItemsControl.ItemsPanel>
+                                                        <ItemsControl.ItemTemplate>
+                                                            <DataTemplate>
+                                                                <Grid>
+                                                                    <Grid.ColumnDefinitions>
+                                                                        <ColumnDefinition Width="20" />
+                                                                        <ColumnDefinition Width="*" />
+                                                                    </Grid.ColumnDefinitions>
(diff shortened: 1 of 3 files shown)


src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -118,7 +118,11 @@ public void ShowSagaWindow()
             {
                 selection.SelectedMessage = SelectedMessage.Message;
             }
-            eventAggregator.PublishOnUIThread(SwitchToSagaWindow.Instance);
+
+            if (selection.SelectedMessage?.Sagas?.Count == 1)
```
mikeminutillo: Can this just rely on `HasSaga`? Do we grey out the tab or something or do we just refuse to display the tab?

src/ServiceInsight/MessageFlow/MessageFlowView.xaml
```diff
@@ -364,32 +364,68 @@
...
+                                                                            <Condition Binding="{Binding IsSagaInitiated}" Value="False" />
+                                                                            <Condition Binding="{Binding IsSagaCompleted}" Value="False" />
+                                                                        </MultiDataTrigger.Conditions>
+                                                                        <Setter TargetName="SagaStatusIcon" Property="Template" Value="{StaticResource SagaInitiatedIcon}" />
+                                                                        <Setter TargetName="SagaStatusIcon" Property="ToolTip" Value="Saga Initiated / Updated" />
+                                                                    </MultiDataTrigger>
+                                                                    <MultiDataTrigger d:DataContext="{d:DesignInstance local:SagaInfo}">
+                                                                        <MultiDataTrigger.Conditions>
+                                                                            <Condition Binding="{Binding IsSagaInitiated}" Value="False" />
```
mikeminutillo: Can a saga be initiated and completed in the same message? If so, wouldn't this show the trigger icon?

src/ServiceInsight/MessageFlow/MessageNode.cs
```diff
@@ -39,20 +48,27 @@ public MessageNode(MessageFlowViewModel owner, StoredMessage message)
...
-                return string.Empty;
+                yield break;
             }
 
-            var originatingSaga = message.Sagas.FirstOrDefault();
-            if (originatingSaga == null)
+            foreach (var saga in message.Sagas)
             {
-                return string.Empty;
+                var originatingSaga = saga;
```
mikeminutillo: Is this still the right name for this variable?

src/ServiceInsight/MessageFlow/MessageNode.cs
```diff
@@ -39,20 +48,27 @@ public MessageNode(MessageFlowViewModel owner, StoredMessage message)
...
+                if (originatingSaga != null)
+                {
+                    yield return new SagaInfo
+                    {
+                        ID = saga.SagaId,
+                        SagaType = TypeHumanizer.ToName(originatingSaga.SagaType),
+                        IsSagaCompleted = saga.ChangeStatus == "Completed",
+                        IsSagaInitiated = string.IsNullOrEmpty(Message.GetHeaderByKey(MessageHeaderKeys.SagaId)) && !string.IsNullOrEmpty(Message.GetHeaderByKey(MessageHeaderKeys.OriginatedSagaId))
```
mikeminutillo: Is it worth pushing the logic for these two down into `SagaInfo`

src/ServiceInsight/MessageFlow/MessageNode.cs
HEskandari: Probably not. I'll rename.

src/ServiceInsight/MessageFlow/MessageNode.cs
HEskandari: yeah, good idea.

src/ServiceInsight/MessageFlow/MessageFlowView.xaml
HEskandari: maybe...but didn't change any logic there :)

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
HEskandari: This added check means if you click on the 'saga' part of the message node in Flow Diagram, it won't take you to the Saga View (if we have more than one saga to show).

You can't rely on HasSaga as if you have more than one, that'd return true.

HEskandari:
@mikeminutillo I have updated the PR with the changes in your comments.

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -118,7 +118,11 @@ public void ShowSagaWindow()
             {
                 selection.SelectedMessage = SelectedMessage.Message;
             }
-            eventAggregator.PublishOnUIThread(SwitchToSagaWindow.Instance);
+
+            if (selection.SelectedMessage?.HasOneSaga == true)
```
mikeminutillo: This is because if you have two sagas we don't know which one to show?

HEskandari: @mikeminutillo Yes. In Flow Diagram, you could click on the saga part and it'd take you to the view showing the incorrect saga. This can be revisited once we have a way of showing multiple sagas in the Saga View.

mikeminutillo: Would it be better to show the first one? The Saga View will still work for it and that keeps the functionality the same as it is today.

HEskandari: @mikeminutillo 	agreed. Rolled back these bits so it is back to its original functionality, although I find it confusing.

PR #838, opened 2019-11-26
Option to disable a splash dialog on Startup

### LINKED ISSUE #806: Option to remove/disable startup splash dialog
The startup dialog is pretty annoying to show. It adds an unneeded startup delay of several seconds.

It would be nice if the dialog could be suppressed and maybe only shown if there is a licensing error.

HEskandari:
Resolves #806 

@aleksandr-samila I have created a PR for you.

### DIFF
--- a/src/ServiceInsight/App.xaml.cs
+++ b/src/ServiceInsight/App.xaml.cs
@@ -8,6 +8,7 @@
     using DevExpress.Xpf.Core;
     using Framework.Logging;
     using ServiceInsight.Framework.Settings;
+    using ServiceInsight.Startup;
     using Shell;
 
     public partial class App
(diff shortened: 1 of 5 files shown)


src/ServiceInsight/App.xaml.cs
```diff
@@ -24,7 +26,14 @@ public App()
         protected override void OnStartup(StartupEventArgs e)
         {
             LogTo.Information("Starting the application...");
-            DXSplashScreen.Show(o => AboutView.AsSplashScreen(), null, null, null);
+
+            var doNotShowSplashScreen = e.Args.Length > 0 && e.Args.Contains ("--silent", StringComparer.OrdinalIgnoreCase);
```
HEskandari: We already have command arg parsing into typed classes. Look at `CommandLineOptions` and `CommandLineArgParser`. 

On Startup, you'll probably need to call:

```
var args = new CommandLineArgParser().Parse();
if(args.SilentStart) ...
```

aleksandr-samila: I saw these classes but decided that they are not currently useful here. And looking deeply today I have the same feelings. E.g. because
`new CommandLineArgParser().Parse();`
implemented to work only with URLs params. 
And I don't see any reason to use `EnvironmentWrapper` that uses `Environment.GetCommandLineArgs()` because we already have an access to startup args via `OnStartup` signature.

Let's talk about this tomorrow during our sync?

aleksandr-samila:
1. Do we need one more reviewer?
2. Do we need to mention about this new startup argument somewhere. E.g. here: https://docs.particular.net/serviceinsight/application-invocation

HEskandari:
@aleksandr-samila PR created to update the doco [here](https://github.com/Particular/docs.particular.net/pull/4614).

Issue #39, opened 2013-10-13
Check for updates


dannycohen:
As Dave, I would like to be notified when a new version of SI is available.
1. When ServiceInsight starts, check for the latest released version of Serviceinsight by calling: http://particular.net/api/products/ServiceInsight/Current
   - Note: since ServiceInsight is not yet released, it has no "current' entry and the above call will result in 404. 
   - For the prerelease versions of Serviceinsight use the following URI: http://particular.net/api/products/ServiceInsight/prerelease
2. Compare the version field number of the current release to the ServiceInsight version installed. 
3. If the current version is newer than the currently installed version, Display the following dialog:

```
A newer version of ServiceInsight [NewVersionNumber] is available for download at: http://particular.net/downloads. 
For more information on version [NewVersionNumber] see https://github.com/Particular/ServiceInsight/releases/tag/[NewVersionNumber] 
```
- Note: for an example of a release info, see: https://github.com/Particular/NServiceBus/releases/tag/4.0.5
- The user can start downloading ServiceInsight immediately by clicking "Download Now" (open the URL: http://particular.net/downloads#DownloadServiceInsight
- The dialog can be discarded using a click on "Cancel"

**Important Notes:**
- The URI calls to check the latest release  must not delay the loading time of ServiceInsight. (It must be done as a background async operation)
- If there is not internet connection or the Particular.net site is not accessible for any reason, do not display an error message
- Log the call to latest release and its result in the Logs pane of ServiceInsight (like you we do with calls to ServiceControl)
- The message dialog must be modeless (it must not interfere with the user's work; it is intended only to draw the user's attention; it can be implemented as a toast popup that can be hidden via a click as stated above).

distantcam:
Coming back to this one...

The platform installer does notify of updates, but isn't always run.

How do people feel about auto-updating like Google Chrome?

johnsimons:
This is a duplicate of https://github.com/Particular/PlatformDevelopment/issues/407, we should close one

PR #840, opened 2019-11-29
Check for updates

aleksandr-samila:
Resolves https://github.com/Particular/ServiceInsight/issues/39


aleksandr-samila:
I created this PR just to follow the common practice.

But before merging I have a few comments:

1.  I wan't able to make this line working: https://github.com/Particular/ServiceInsight/blob/check-for-updates/src/ServiceInsight/Shell/UpdateManager.cs#L32. It doesn't fail but there is nothing in the Log tab.
2. https://github.com/Particular/ServiceInsight/blob/check-for-updates/src/ServiceInsight/Shell/UpdateManager.cs#L34-L35 just for test. Because running SI locally - current app version is e.g. `1.13.1-check-for-updates.1` and can't be used to create a new `Version` class.
3. Currently the result is:
image

HEskandari:
@aleksandr-samila applied some changes:

- Moved the UI component to the right-hand side
- Added an icon
- Slightly changed the text message
- Exposed the version checker as an interface and registered in the IoC container
- Parsed the version suffixed when checking for updated version

PR #852, opened 2020-03-31
GitHubSync update

particularbot:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


bording:
@andreasohlund @ramonsmits RepoSync shouldn't be enabled on this repo.

andreasohlund:
Why not? It has gitignores, license, contributing etc?

Do we need to make the sync more selective?

PR #853, opened 2020-04-03
Reposync ignore

ramonsmits:
Close https://github.com/Particular/ServiceInsight/pull/852 after merge


andreasohlund:
Not sure about this one, this is an active repo, I think we should just ignore needed files?

PR #856, opened 2020-04-23
Update license messages

mikeminutillo:
Updates the license messages to match the new licensing model.


HEskandari:
Code-wise I think it is okay, do I need to do some testing before it is merged in?

PR #864, opened 2020-07-02
Updating project format

HEskandari:
Use the latest SDK style .csproj format and upgrade to .NET 4.8


src/ServiceInsight.Licensing/InternalVisibleTo.cs
```diff
@@ -0,0 +1,4 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("ServiceInsight")]
```
mikeminutillo, line 3: Why is this necessary? Is this just to prevent us from making the licensing bits public? Is that as important in a product?

src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
```diff
@@ -16,7 +16,7 @@
                 <ResourceDictionary>
                     <SolidColorBrush x:Key="GlyphBrush" Color="#444" />
                     <SolidColorBrush x:Key="DisabledForegroundBrush" Color="#888" />
-                    <SolidColorBrush x:Key="WindowBackgroundBrush" Color="#000" />
+                    <!-- <SolidColorBrush x:Key="WindowBackgroundBrush" Color="#000" /> -->
```
mikeminutillo, line 19: If this is not used, can it be removed?

src/ServiceInsight/ServiceInsight.csproj
```diff
@@ -1,941 +1,94 @@
...
-    <ErrorReport>prompt</ErrorReport>
-    <Prefer32Bit>false</Prefer32Bit>
-    <UseVSHostingProcess>false</UseVSHostingProcess>
-    <CodeAnalysisRuleSet>..\CommonRuleSet.ruleset</CodeAnalysisRuleSet>
-  </PropertyGroup>
-  <PropertyGroup>
+    <Company>NServiceBus Ltd.</Company>
     <ApplicationIcon>ServiceInsight.ico</ApplicationIcon>
+    <UseWpf>true</UseWpf>
+    <DefineConstants>$(DefineConstants);REGISTRYLICENSESOURCE</DefineConstants>
```
mikeminutillo, line 9: Is this still needed for the main EXE?

src/ServiceInsight.Licensing/InternalVisibleTo.cs
HEskandari: This is important because we didn't want to make some of our licensing code public. We can't just make part of it public in SI codebase either, since it relies on some other 'internal' bits from the Licensing packages.

src/ServiceInsight/ServiceInsight.csproj
HEskandari: Probably not. I'll remove and verify.

HEskandari: removed

src/ServiceInsight/MessageFlow/MessageActionPopup.xaml
HEskandari: tested and removed

PR #867, opened 2020-07-13
change the logging source

### LINKED ISSUE #866: LogWindow does not show any log entries
## Who's affected

This affects all users of ServiceInsight version 2.0. 

## Symptoms

The log window does not show any log entries of ServiceControl communication and it remains empty.

HEskandari:
Closes #866


HEskandari:
@mikeminutillo would you be able to have a look at this?

PR #873, opened 2020-09-01
Display GUIDs values in Saga diagram

### LINKED ISSUE #811: Saga view shows Guid values as "36 bytes"
Shouldn't it be possible to show the actual value for a Guid (or any value) type?

Observed when @indualagarsamy gave her saga talk at Explore DDD.

HEskandari:
Closes #811


seanfarmar (approved):
LGTM

PR #883, opened 2020-09-03
Implemented single instance check #250

### LINKED ISSUE #250: ServiceInsight should be single instance
If the system is already open, it shouldn't open again.

//cc @udidahan

seanfarmar:
Resolves #250 
Used this a guide: https://www.c-sharpcorner.com/blogs/how-to-run-single-instance-of-your-application


src/ServiceInsight/App.xaml.cs
```diff
@@ -25,6 +26,9 @@ public App()
         protected override void OnStartup(StartupEventArgs e)
         {
             LogTo.Information("Starting the application...");
+                    
+            if (SingleInstance.AlreadyRunning())
```
SimonCropp, line 30: why not use a mutex?
```
    using var mutex = new Mutex(true, "ServiceInsight", out var createdNew);
        if (!createdNew)
        {
               App.Current.Shutdown();
        } 
```

HEskandari: @SimonCropp we decided to go with a mix of Win32 and mutex. There is no native way to bring the window into foreground when you launch a new instance.

PR #885, opened 2020-09-04
Show vector icons in Messages list view

### LINKED ISSUE #504: ServiceInsight looks blurry on higher DPI screens
I just noticed this on my retina display. SI looks blurry compared to other elements on the screen.

https://msdn.microsoft.com/en-us/library/windows/desktop/ee308410(v=vs.85).aspx

SI should be updated to be per-monitor DPI aware. There might be other things impacted as we use bitmaps in places instead of scalable graphics.

HEskandari:
Closes #504


seanfarmar (approved):
LGTM

PR #894, opened 2020-09-09
2nd iteration of 881

### LINKED ISSUE #881: When starting SI and sorting data by a unsupported column, no data is displayed
Related to #880 

The core of the issue is that the Messages window in SI has some columns that ServiceControl does not support sorting on. ServiceInsight does client-side sorting. When there is no data to sort (for example cold run) and the last sort column is on one of these unsupported columns, no data is loaded as ServiceControl returns a 500 status (Internal Server Error). 

The lack of error visibility makes this harder to understand and user wouldn't know the sorting is the issue.

A potential solution is to fall back to a sort-supported column (e.g. at start) or apply a default sort (e.g. message ID).

seanfarmar:
resolves #881 
2nd iteration of 881

* Adding tool tips
* Make the column headers un-clickable


src/ServiceInsight/MessageList/MessageListView.xaml
```diff
@@ -174,7 +174,7 @@
                                     CellTemplate="{StaticResource TextEditGridColumnCellTemplate}"
                                     FieldName="ProcessingEnded"
                                     Header="Processing Ended"
-                                    HeaderToolTip="When the processing ended"
+                                    HeaderToolTip="When the processing ende - Sorting not supportedd"
```
HEskandari, line 177:
```suggestion
                                    HeaderToolTip="When the processing end - Sorting not supportedd"
```

HEskandari (approved):

PR #902, opened 2020-09-18
Add Particular.CodeRules 0.4.0

kbaley:


src/ServiceInsight.Tests/AboutViewModelTests.cs
```diff
@@ -34,7 +34,7 @@ public void Should_fetch_service_control_version()
         {
             ((IActivate)sut).Activate();
 
-            serviceControl.Received(1).GetVersion();
+            _ = serviceControl.Received(1).GetVersion();
```
adamralph, line 37: This won't fix this test. This test was never proving anything, since the call to `GetVersion` is fire-and-forget. The correct fix is:

```C#
        [Test]
        public async Task Should_fetch_service_control_version()
        {
            ((IActivate)sut).Activate();

            await serviceControl.Received(1).GetVersion();
        }
```

or

```C#
        [Test]
        public Task Should_fetch_service_control_version()
        {
            ((IActivate)sut).Activate();

            return serviceControl.Received(1).GetVersion();
        }
```

I'll push a commit which fixes these.

adamralph:
@kbaley I pushed a few more commits. Can you please take a look?

If you agree with them, I think we should squash all these commits into one before merging. I don't think any of them have independent value. Arguably, the R# removal could be split from the addition of CodeRules, but I don't think it's important.

kbaley:
@adamralph These look good to me. I can't merge without an approving review and I can't approve my own PR so go ahead and squash these and merge when you can.

adamralph (approved):

PR #942, opened 2020-10-28
downgrading to net461

HEskandari:


HEskandari:
Not required anymore.

Issue #941, opened 2020-10-27
ServiceInsight SingleInstance mode does not pass the invocation parameters to the running instance


HEskandari:
Otherwise, the single instance mode would break the SP -> SI workflow as once the instance is running, clicking a link in SP would only show the currently running application window without taking into account the new context. The second instance might have different invocation parameters (clicking a link on a different message from ServicePulse would have things like search queries or connection to a different instance of ServiceControl) so the parameters should be taken into account when SI is already opened.

The problem with that is, that the two instances run in separate processes, which means some sort of cross-process communication is needed to pass the parameters (IPC channel or TCP listener).

HEskandari:
Resolved by #974

PR #974, opened 2020-12-07
Sending the command line arguments to the singleton instance

HEskandari:
_Connects to https://github.com/Particular/ServiceInsight/issues/941_

Uses a TcpClient to send the command args to an existing instance when second instance launches.

### DIFF
--- a/src/ServiceInsight.Tests/ShellViewModelTests.cs
+++ b/src/ServiceInsight.Tests/ShellViewModelTests.cs
@@ -24,6 +24,7 @@
     using ServiceInsight.Shell;
     using ServiceInsight.Startup;
     using Shouldly;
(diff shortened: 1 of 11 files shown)


mauroservienti:
My only concern with using TCP is that if the port s in use by another application the user would need to manually find an available one and configure it; which is problematic for multiple reasons. Another approach could be to use a named pipe, that can be named using a GUID (for example the same used by the installer as product identifier) so that it's unique and prevents any possible collision. What do you think?

HEskandari:
@mauroservienti named pipes are similar in functionality. I'll see if I can swap it.

HEskandari:
@mauroservienti updated.

mauroservienti (approved):

PR #976, opened 2020-12-14
fix compatibility with running older versions running at the same time

HEskandari:
Connects to #941

Fix a compatibility issue when passing command line arguments over to older versions (with no named-pipe support).


mauroservienti (approved):

mauroservienti:
Is the result of this that the current instance will be used to show details instead of the already running one that is probably an older version?

HEskandari:
@mauroservienti this came up when I was trying to launch a newer version of SI, while an older version was already running. The new instance can not connect via named-pipes, as the old versions do not support it, which caused an Timeout exception and the new instance would terminate. 

Technically it was still working correctly, but this would log those type of errors as additional measure.

PR #985, opened 2020-12-18
Making a change to get a message body for audited messages

WojcikMike:
Ok, this PR seems to solve the problem of bodies. I had to leave the MessageBodyEndpoint const as this works for error messages. I tested it for audits and errors and it seems to be ok.

### DIFF
--- a/src/ServiceInsight/Saga/SagaMessage.cs
+++ b/src/ServiceInsight/Saga/SagaMessage.cs
@@ -20,6 +20,8 @@
     {
         public string MessageId { get; set; }
 
+        public string BodyUrl { get; set; }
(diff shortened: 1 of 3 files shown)


mikeminutillo:
This was throwing an exception for me when I load my test data. I tried to figure out why/how the message was getting added to the diagram without the needed detail, but I ended up just pushing the query down into the ServiceControl layer in https://github.com/Particular/ServiceInsight/pull/986

This seems to resolve the problem for me and now saga data is being loaded properly.

HEskandari (approved):

mikeminutillo (approved):

PR #986, opened 2020-12-21
Push query down to SC access layer

mikeminutillo:
Pushes the query from https://github.com/Particular/ServiceInsight/pull/985 down to the ServiceControl layer. It is not possible to get saga message data without also retrieving the body uls.


WojcikMike (approved):

mikeminutillo:
Made a small tweak to get the original working https://github.com/Particular/ServiceInsight/pull/985

PR #996, opened 2021-01-20
Read license from PARTICULARSOFTWARE_LICENSE environment variable

mikeminutillo:
This change allows ServiceInsight to read a license stored in the PARTICULARSOFTWARE_LICENSE environment variable.

### DIFF
--- a/src/ServiceInsight.Licensing/ServiceInsight.Licensing.csproj
+++ b/src/ServiceInsight.Licensing/ServiceInsight.Licensing.csproj
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Particular.CodeRules" Version="0.7.0" PrivateAssets="All" />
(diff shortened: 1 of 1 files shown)


HEskandari:
Wouldn't that make it more confusing? imagine having an outdated licenses, somewhere EnvVar is pointing to. With SI being an app, I'd say more explicit and limited license location would give a better user experience (e.g. application or app data folder)

mikeminutillo:
The licensing code will not select an out of date license. It scans all available locations and selects the best available that is valid for ServiceInsight.

To provide the alternative perspective, imagine that your endpoints and local ServiceControl installation all rely on the same environment variable, but ServiceInsight ignores that and claims to be unlicensed.

mikeminutillo:
Handled in https://github.com/Particular/ServiceInsight/pull/1010

PR #997, opened 2021-01-22
Remove separate licensing project

bording:
The .NET 5.0.102 SDK has a new `IncludePackageReferencesDuringMarkupCompilation` setting that lets us get rid of the separate licensing project.

This also includes some other misc project file cleanups, and ensure that's that all shipped assemblies will be versioned by moving the Particular.Packaging reference into a props file.


bording:
And of course it's failing because 5.0.102 hasn't been installed on the build agents yet...

HEskandari (approved):

Issue #1042, opened 2021-04-21
Connecting to ServiceControl from a redirected URL fails


HEskandari:
## Who's affected

Customers connecting to a ServiceControl instance via a redirected URL are affected by this bug.

## Symptoms

When connecting to a ServiceControl instance via a redirected URL, ServiceInsight does accept the connection, but fails to retrieve messages with the following exception:

```
System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at ServiceInsight.ServiceControl.DefaultServiceControl.<>c__DisplayClass39_0`1.<GetPagedResult>b__0(IRestResponse`1 response) in /_/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs:line 306
   at ServiceInsight.ServiceControl.DefaultServiceControl.ProcessResponse[T,T2](Func`2 selector, IRestResponse`1 response) in /_/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs:line 604
   at ServiceInsight.ServiceControl.DefaultServiceControl.<Execute>d__42`2.MoveNext() in /_/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs:line 531
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at ServiceInsight.MessageList.MessageListViewModel.<RefreshMessages>d__46.MoveNext() in /_/src/ServiceInsight/MessageList/MessageListViewModel.cs:line 163
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at ServiceInsight.MessageList.MessageListViewModel.<RefreshMessages>d__45.MoveNext() in /_/src/ServiceInsight/MessageList/MessageListViewModel.cs:line 149
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1043, opened 2021-04-21
support redirect when connecting to ServiceControl instances

HEskandari:
Connects to https://github.com/Particular/ServiceInsight/issues/1042

### DIFF
--- a/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
+++ b/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
@@ -168,11 +168,12 @@
             }
 
             var available = false;
-            using (workNotifier.NotifyOfWork($"Verifying ServiceControl availability at {url}"))
+            var address = url;
+
+            using (workNotifier.NotifyOfWork($"Verifying ServiceControl availability at {address}"))
             {
-                clientRegistry.EnsureServiceControlClient(url);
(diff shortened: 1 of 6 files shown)


src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
```diff
@@ -168,21 +168,24 @@ public async Task ConnectToService(string url)
...
                 return;
             }
 
-            using (workNotifier.NotifyOfWork($"Connecting to ServiceControl at {url}"))
+            clientRegistry.EnsureServiceControlClient(address);
+
+            using (workNotifier.NotifyOfWork($"Connecting to ServiceControl at {address}"))
             {
-                var node = AddServiceNode(url);
+                var node = AddServiceNode(address);
```
mikeminutillo, line 188: Does this mean that the customer is going to see the redirected url in their list of saved options and not the one they entered? It's not necessarily an issue, I just wanted to make sure it is intended..

mikeminutillo (approved):

src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
HEskandari: yes, that's how it'll be. This will treat the two connections as one.

Issue #1033, opened 2021-03-25
ServiceInsight does not allow running multiple instances on the same machine via remote desktop


kbaley:
## Reproduction steps

- Have ServiceInsight 2.6.0 installed and configured to connect to an instance of ServiceControl
- User A connects to the machine via remote desktop and runs ServiceInsight
- ServiceInsight behavior is good/normal
- User B connects to the same machine via remote desktop while User A is still running ServiceInsight and User B now also runs ServiceInsight
- User B’s ServiceInsight will crash instantly or after clicking on endpoint(s) in endpoint explorer
- Users can close the app, and User B can run the app first
- User B will see normal behavior
- User A runs ServiceInsight with User B still using it
- Now User A will crash instantly or after clicking endpoint(s) in endpoint explorer

This was likely introduced in version 2.4 with [this issue](https://github.com/Particular/ServiceInsight/issues/941) as it is working in ServiceInsight 2.3.

kbaley:
One possible fix for this is to incorporate the username into the name of the mutex that's used to check if ServiceInsight is running:

https://github.com/Particular/ServiceInsight/blob/master/src/ServiceInsight/App.xaml.cs#L20

odelljl:
We have this exact issue and are holding on version 2.3 as a result.

HEskandari:
There are two aspects to this:

- The code uses a constant as a mutext to avoid running multi-instance. When running more than one user (via RDP), this causes the second instance to not run. 
- Even in a single user scenario, there's a named pipe communication between second -> first instance, so that we can send the information back to the instance already running. For example, user has a session open already, then clicks the 'Open in ServiceInsight' button in ServicePulse. The information should appear in the instance already running (e.g. message getting selected).

The named pipe and the mutex name (both named 'ServiceInsight') need to be changed to something that allow multiple-user to use it at the same time (e.g. include the logged-in username?)

HEskandari:
More investigation showed that the issue is not the Mutex, as in terminal model (RDP) the mutex would have local visibility which makes it invisible to the second RDP session. The actual underlying issue is an exception that happens due to the second NamedPipe being created on the machine when launching the second instance. The second named pipe would use the same name (ServiceInsight) which would conflict with the first running instance and the application crashes. The exception in the log file was like this:

```
2021-05-27 23:07:36.971 +00:00 [Error] (ServiceInsight.Framework.ExceptionHandler) An unhandled exception occurred. Error message is "All pipe instances are busy.
[...]

PR #1064, opened 2021-05-27
Scope mutex and named pipe to the current user

HEskandari:
_Resolves https://github.com/Particular/ServiceInsight/issues/1033_

- Scopes mutex to the local user/session
- Appends current username to the named pipe to make it distinct per user session

### DIFF
--- a/src/ServiceInsight/Startup/StartupConfigListener.cs
+++ b/src/ServiceInsight/Startup/StartupConfigListener.cs
@@ -40,7 +40,7 @@
 
         static async Task Handle(IEventAggregator eventAggregator, CommandLineArgParser parser, CancellationToken cancellationToken)
         {
-            using (var pipe = new NamedPipeServerStream("ServiceInsight", PipeDirection.In, 1, PipeTransmissionMode.Byte))
+            using (var pipe = new NamedPipeServerStream($"ServiceInsight-{Environment.UserName}", PipeDirection.In, 1, PipeTransmissionMode.Byte))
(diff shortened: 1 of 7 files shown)


SeanFeldman (approved):

SeanFeldman (approved):

src/ServiceInsight/Startup/StartupConfigListener.cs
```diff
@@ -40,7 +40,7 @@ public static async Task Start(IEventAggregator eventAggregator, CommandLineArgP
 
         static async Task Handle(IEventAggregator eventAggregator, CommandLineArgParser parser, CancellationToken cancellationToken)
         {
-            using (var pipe = new NamedPipeServerStream("ServiceInsight", PipeDirection.In, 1, PipeTransmissionMode.Byte))
+            using (var pipe = new NamedPipeServerStream($"ServiceInsight-{Environment.UserName}", PipeDirection.In, 1, PipeTransmissionMode.Byte))
```
SeanFeldman, line 43: nit: I'm not a big fan of this duplication in two different files. We know the name has to be identical in two places; therefore, we could extract it into a component/static class and consolidate it.

HEskandari: Agreed. I'll update it.

SeanFeldman: I took care of it in https://github.com/Particular/ServiceInsight/pull/1064/commits/ad2b694f0243a227146188af91366899203eddc2

PR #1157, opened 2021-12-31
Fix Successful with Warn icon

### LINKED ISSUE #1156: Icon for "Successfully Processed with Warning" is incorrect
The icon that is used for Successfully Processed with Warning looks exactly like the Resolved Successfully icon

It should look like this:

image

But it looks like this:

image

Here it is in the UI:

mikeminutillo:
Fixes #1156

The `MessageStatus_Successful_Warn` icon was a copy of Successfully Retried. This change sets it back to being a copy of Successful with the warning overlay.

Before:

image

After:

image


HEskandari (approved):

HEskandari:
@mikeminutillo can this be merged?

PR #1171, opened 2022-02-09
Allows linking nodes in the message flow view as a fallback when no link could be established on OriginatingEndpoint / ProcessingEndpoint mismatch

ramonsmits:
- Resolves: https://github.com/Particular/ServiceInsight/issues/1196

In order to link nodes the following must be true:

```c#
child.RelatedTo == parent.Id && child.OriginatingEndpoint == parent.ProcessingEndpoint
```    

However, in situaties where only processed messages can be linked based on RelatedTo these messages remain unlinked in the message flow view.

Before:

image

After:

image

Logic is written in such a way that the fallback only happens if intent not equal to publish. Meaning, only messages and commands as these SHOULD only have one processing endpoint

## Implications

This is a fallback, the default linking behavior remains the same is the message flow  is purely passing only NServiceBus endpoints but adds linking when that is not always the case (i.e. native integration).

### DIFF
--- a/src/ServiceInsight/Framework/Logging/LoggingConfig.cs
+++ b/src/ServiceInsight/Framework/Logging/LoggingConfig.cs
@@ -13,6 +13,7 @@
     using Serilog.Events;
     using Serilog.Filters;
     using ServiceControl;
+    using ServiceInsight.MessageFlow;
 
     public static class LoggingConfig
     {
@@ -31,7 +32,7 @@
                 .WriteTo.Trace(outputTemplate: "[{Level}] ({SourceContext}) {Message}{NewLine}{Exception}")
                 .WriteTo.Logger(lc => lc
                     .MinimumLevel.Verbose()
-                    .Filter.ByIncludingOnly(MatchingTypes(typeof(DefaultServiceControl), typeof(CustomMessageViewerResolver)))
+                    .Filter.ByIncludingOnly(MatchingTypes(typeof(DefaultServiceControl), typeof(CustomMessageViewerResolver), typeof(MessageFlowViewModel)))
                     .WriteTo.Observers(logEvents => logEvents
                         .ObserveOn(TaskPoolScheduler.Default)
                         .Do(LogWindowViewModel.LogObserver)
--- a/src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
+++ b/src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
@@ -268,9 +268,25 @@
                 // [CM] I don't know how it's happening, but a user reported an
                 // error where multiple results were returned from this query.
                 var parentMessages = nodeMap.Values.Where(m =>
-                    m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null &&
-                    m.Message.MessageId == msg.Message.RelatedToMessageId &&
-                    m.Message.ReceivingEndpoint.Name == msg.Message.SendingEndpoint.Name);
+                    m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null
+                    && m.Message.MessageId == msg.Message.RelatedToMessageId
+                    && m.Message.ReceivingEndpoint.Name == msg.Message.SendingEndpoint.Name // Needed with publishes as identical events can be consumed by multiple endpoints
+                    );
+
+                // Fallback, get "parent" when originating message is not an event (publish)
+                if (!parentMessages.Any())
+                {
+                    // TODO: Log INFO indicating no parent match on "endpoint", only on message identifier.
+                    parentMessages = nodeMap.Values.Where(m =>
+                        m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null &&
+                        m.Message.MessageId == msg.Message.RelatedToMessageId && m.Message.MessageIntent != MessageIntent.Publish
+                        );
+                }
+
+                if (!parentMessages.Any())
+                {
+                    // TODO: Log WARNING to inform user
+                }
 
                 foreach (var parentMessage in parentMessages)
                 {
--- a/src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
+++ b/src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
@@ -287,11 +287,11 @@
 
                 if (!parentMessages.Any())
                 {
-                    LogTo.Error("No parent could be resolved for message {0} which has RelatedToMessageId set. This can happen if the parent has been purged due to retention expiration, an ServiceControl node to be unavailable, or because the parent message not been stored (yet).", msg.Message.MessageId);
+                    LogTo.Information("No parent could be resolved for the message with Id '{0}' which has RelatedToMessageId set. This can happen if the parent has been purged due to retention expiration, an ServiceControl node to be unavailable, or because the parent message not been stored (yet).", msg.Message.MessageId);
                 }
                 else if (parentMessages.Count() > 1)
                 {
-                    LogTo.Error("Multiple parents matched for {0} possibly due to more-than-once processing, linking to all as unknown which processing attemps generated the message.", msg.Message.MessageId);
+                    LogTo.Error("Multiple parents matched for message id '{0}' possibly due to more-than-once processing, linking to all as it is unknown which processing attempt generated the message.", msg.Message.MessageId);
                 }
 
                 foreach (var parentMessage in parentMessages)
--- a/src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
+++ b/src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
@@ -12,6 +12,7 @@
     using Caliburn.Micro;
     using Mindscape.WpfDiagramming;
     using Mindscape.WpfDiagramming.FlowDiagrams;
+    using Anotar.Serilog;
     using ServiceInsight.Models;
     using ServiceInsight.ServiceControl;
     using ServiceInsight.Framework;
@@ -259,14 +260,11 @@
         {
             foreach (var msg in relatedMessagesTask)
             {
-                if (msg.Message.RelatedToMessageId == null &&
-                    msg.Message.RelatedToMessageId != msg.Message.MessageId)
+                if (string.IsNullOrEmpty(msg.Message.RelatedToMessageId) && msg.Message.RelatedToMessageId != msg.Message.MessageId)
                 {
                     continue;
                 }
 
-                // [CM] I don't know how it's happening, but a user reported an
-                // error where multiple results were returned from this query.
                 var parentMessages = nodeMap.Values.Where(m =>
                     m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null
                     && m.Message.MessageId == msg.Message.RelatedToMessageId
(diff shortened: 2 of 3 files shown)


src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -268,9 +268,25 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
+                    // TODO: Log INFO indicating no parent match on "endpoint", only on message identifier.
+                    parentMessages = nodeMap.Values.Where(m =>
+                        m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null &&
+                        m.Message.MessageId == msg.Message.RelatedToMessageId && m.Message.MessageIntent != MessageIntent.Publish
+                        );
+                }
+
+                if (!parentMessages.Any())
+                {
+                    // TODO: Log WARNING to inform user
```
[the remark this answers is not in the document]
ramonsmits: Added in #ef245aae

ramonsmits:
> Do you think it's possible to add a unit test for this?

@WilliamBZA No tests exist at all against this view model. Potentially the `LinkConversationNodes` method could be extracted and unit tested as the input/output is a conversion.

WilliamBZA:
I'm not sure that I'm the right person to review this PR. I can review the code but am not really sure I'm able to follow the implication of the change.

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -259,18 +260,39 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
+
+                    if (parentMessages.Any())
+                    {
+                        LogTo.Warning("Fall back to match only on RelatedToMessageId for message {0} matched but link could be invalid.", msg.Message.MessageId);
+                    }
+                }
+
+                if (!parentMessages.Any())
+                {
+                    LogTo.Error("No parent could be resolved for message {0} which has RelatedToMessageId set. This can happen if the parent has been purged due to retention expiration, an ServiceControl node to be unavailable, or because the parent message not been stored (yet).", msg.Message.MessageId);
```
danielmarbach, line 290: Given that these log statements indicate this can happen and is expected, wouldn't an info or even debug level be enough?

danielmarbach, line 290:
```suggestion
                    LogTo.Error("No parent could be resolved for the message with Id '{0}' which has RelatedToMessageId set. This can happen if the parent has been purged due to retention expiration, an ServiceControl node to be unavailable, or because the parent message not been stored (yet).", msg.Message.MessageId);
```

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -259,18 +260,39 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
+                }
+                else if (parentMessages.Count() > 1)
+                {
+                    LogTo.Error("Multiple parents matched for {0} possibly due to more-than-once processing, linking to all as unknown which processing attemps generated the message.", msg.Message.MessageId);
```
danielmarbach, line 294:
```suggestion
                    LogTo.Error("Multiple parents matched for message id '{0}' possibly due to more-than-once processing, linking to all as unknown which processing attempts generated the message.", msg.Message.MessageId);
```

danielmarbach, line 294: Where are errors logged to? Should we tone this down a bit? What would a user do with an error and why should they care?

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -259,18 +260,39 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
+                if (!parentMessages.Any())
+                {
+                    parentMessages = nodeMap.Values.Where(m =>
+                        m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null &&
+                        m.Message.MessageId == msg.Message.RelatedToMessageId && m.Message.MessageIntent != MessageIntent.Publish
+                        );
+
+                    if (parentMessages.Any())
+                    {
+                        LogTo.Warning("Fall back to match only on RelatedToMessageId for message {0} matched but link could be invalid.", msg.Message.MessageId);
```
danielmarbach, line 284: Again what is that a warning? How is this information important to be a warning when nothing can be done from the user perspective?

danielmarbach (commented):
I don't have enough context to argue whether we should proceed with this. I did try to provide some nitpicks though :trollface:

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
adamralph: Agree. An error conveys "something is wrong - the application cannot function as intended".

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
adamralph: ditto https://github.com/Particular/ServiceInsight/pull/1171#discussion_r826716375

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
adamralph: Agree. Similar to https://github.com/Particular/ServiceInsight/pull/1171#discussion_r826716375, a warning conveys "something _may be_ wrong - the application _may_ not function as intended".

HEskandari:
@danielmarbach to answer your question: logs are displayed within SI in the Logs window. The reason to mark them as Warn/Error was to make them stand out as we log other things there as well. If you want to argue "what's the action" you can argue why we log anything at all.

The error logs in the diagram code are not 'expected'. Maybe those can be replaced by exceptions. Those conditions are not expected to happen. When customers do report those, we'll need to check their (SC) data to see why that happens.

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
ramonsmits: @danielmarbach its a missing gap in our platform. 

- https://github.com/Particular/NServiceBus/pull/6264

Currently no correlation is done and this at least tries to do a "best effort" that still can be wrong hence the warning to inform the user about this detected state.

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
ramonsmits: > Given that these log statements indicate this can happen and is expected, wouldn't an info or even debug level be enough?

In an ideal world this would be visualized, showing something like a question mark box in the diagram with information to explain the visual state. I personally think showing an error if we detect that a graph is likely incomplete isn't bad but I can also understand your reasoning. I'll modify your code suggest.

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
ramonsmits: @danielmarbach Logs are outputted in SI in the logs tab: https://docs.particular.net/serviceinsight/logging

At least the error/warning indicates that the data visualized is correct or in this case an explanation for the strange multiple connections. Yes, it's not ideal but it's better than nothing.

WilliamBZA (approved):

ramonsmits:
Is rebased on top of dev-exp-upgrade-buid-fix, needs https://github.com/Particular/ServiceInsight/pull/1288 to be merged first.

src/ServiceInsight/Framework/Logging/LoggingConfig.cs
```diff
@@ -31,7 +32,7 @@ public static void SetupLogging()
                 .WriteTo.Trace(outputTemplate: "[{Level}] ({SourceContext}) {Message}{NewLine}{Exception}")
                 .WriteTo.Logger(lc => lc
                     .MinimumLevel.Verbose()
-                    .Filter.ByIncludingOnly(MatchingTypes(typeof(DefaultServiceControl), typeof(CustomMessageViewerResolver)))
+                    .Filter.ByIncludingOnly(MatchingTypes(typeof(DefaultServiceControl), typeof(CustomMessageViewerResolver), typeof(MessageFlowViewModel)))
```
HEskandari, line 35: Is this really necessary? Why do you need to customize logs for that view model?

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -259,18 +260,39 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
                     continue;
                 }
 
-                // [CM] I don't know how it's happening, but a user reported an
-                // error where multiple results were returned from this query.
                 var parentMessages = nodeMap.Values.Where(m =>
-                    m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null &&
-                    m.Message.MessageId == msg.Message.RelatedToMessageId &&
-                    m.Message.ReceivingEndpoint.Name == msg.Message.SendingEndpoint.Name);
+                    m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null
```
HEskandari, line 269: The code potentially multi-iterates the IEnumerable. Maybe convert it to the list here? and update th usages e.g. `.Count()` to `.Count` etc.?

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -259,18 +260,39 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
+
+                    if (parentMessages.Any())
+                    {
+                        LogTo.Warning("Fall back to match only on RelatedToMessageId for message {0} matched but link could be invalid.", msg.Message.MessageId);
+                    }
+                }
+
+                if (!parentMessages.Any())
+                {
+                    LogTo.Information("No parent could be resolved for the message with Id '{0}' which has RelatedToMessageId set. This can happen if the parent has been purged due to retention expiration, an ServiceControl node to be unavailable, or because the parent message not been stored (yet).", msg.Message.MessageId);
```
HEskandari, line 290: I think Serilog message templates should have the variable name and not the element number. Did you check if this works? I think it should be:

```
var msg = 1234;
LogTo.Information("....message Id {MessageID}...", msg);
```

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
```diff
@@ -259,18 +260,39 @@ void LinkConversationNodes(IEnumerable<MessageNode> relatedMessagesTask)
...
-                // [CM] I don't know how it's happening, but a user reported an
-                // error where multiple results were returned from this query.
                 var parentMessages = nodeMap.Values.Where(m =>
-                    m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null &&
-                    m.Message.MessageId == msg.Message.RelatedToMessageId &&
-                    m.Message.ReceivingEndpoint.Name == msg.Message.SendingEndpoint.Name);
+                    m.Message != null && m.Message.ReceivingEndpoint != null && m.Message.SendingEndpoint != null
+                    && m.Message.MessageId == msg.Message.RelatedToMessageId
+                    && m.Message.ReceivingEndpoint.Name == msg.Message.SendingEndpoint.Name // Needed with publishes as identical events can be consumed by multiple endpoints
+                    );
```
ramonsmits, line 272:
```suggestion
                    )
                    .ToArray();
```

src/ServiceInsight/MessageFlow/MessageFlowViewModel.cs
ramonsmits: Resolved via 7755458168c0a8d40dfdc8c766cb061eb0dd8cc6

HEskandari (approved):

PR #1189, opened 2022-03-28
Switch to using MinVer instead of GitVersion

bording:


kentdr (commented):
Setup.csproj has two GitVersion references in it still.

kentdr (approved):

PR #1206, opened 2022-05-25
Disabling an option to multiselect rows on the grid

WojcikMike:
The issue was raised: one can select multiple rows on the message list and click retry, but only one message is retried. Currently, mutli-selection on the grid is not handled in any way. Message selected on the list is displayed in the details view (flow diagram, sequence, headers, body), without changing the way how the application work it would be hard to make the multi-select behavior to work in a reasonable way.


mauroservienti (approved):
Verified that it works as expected

PR #1209, opened 2022-05-30
Check if a connection to a ServiceControl instance is already present

### LINKED ISSUE #990: ServiceInsight does not check for duplicate connections at startup
When a second instance of SI is executed, passing startup parameters such as connection to SC to the first instance, the connection need to be de-duplicated (check for existing connection):

- When connection doesn't exist, the connection should be added to endpoint Explorer.
- When the connection does exist, the existing connection/endpoint need to get selected in the Explorer window.

When invoking `View in ServiceInsight` functionality from ServicePulse, the right message need to be displayed (as it currently is).

mauroservienti:
- Fix #990 

Before adding a new connection, the endpoint explorer checks if any of the ServiceControl nodes already connected matches the unique URL. If that's the case, it skips adding the new connection.

### DIFF
--- a/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
+++ b/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
@@ -160,9 +160,19 @@
             }
         }
 
+        bool IsConnectedToServiceControlNode(string url)
+        {
+            return Items.OfType<ServiceControlExplorerItem>().Any(ei => ei.Url == url);
+        }
+
         public async Task ConnectToService(string url)
         {
             if (!url.IsValidUrl())
+            {
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


WojcikMike (approved):

mauroservienti:
This is getting more complex than expected. When the code detects that the SC instance node is already present (SI is already connected to the required SC) it is supposed to select the correct SC node. The selection works, but the search not always...which is already weird by itself. Worse, the explorer tree view doesn't reflect the selection change and it starts behaving in weird ways. I understood why the selection change is not reflected, but fixing it is getting more complex than needed. All thanks to how weird the WPF TreeView control is 😵‍💫

Then there is the mentioned search intermittent behavior Which I haven't investigated yet.

mauroservienti:
Let's consider this as done, it fixes the connection de-duplication issue. I'll raise two follow-up issues for the two identified bugs.

PR #1213, opened 2022-06-01
Updated DevExpress WPF to 21.2

ramonsmits:
Updated to 21.2 to check if it resolved For https://github.com/Particular/KeepTheLightsOn/issues/782 which it did not but pushed just in case we might still update for other reasons.


ramonsmits:
Closing, this did complete but there is no reason to update as this requires a new license.

PR #1214, opened 2022-06-01
Spike: Simple main window with grid with databinding to reproduce "row selection bug"

ramonsmits:


ramonsmits:
Thanks for diagnosing the source @HEskandari ! Fix is in:

- https://github.com/Particular/ServiceInsight/pull/1215

PR #1215, opened 2022-06-02
Messages window no longer multiple rows as selected as user scrolls through data

### LINKED ISSUE #882: Messages window show 'ghost' rows as selected as user scrolls through data
When a row is selected in the Messages window, and then the user scrolls further down the list (without clicking on any other row), you can see that the same row index will appear as selected as it scrolls into the view. 

For example, if the 3rd row from the top is selected, as you scroll the 3rd row out of the view and scroll other rows from the bottom of the table, the 3rd row that appears is also shown as selected. The selection is however not applied or changed since the row is not really selected, but it appears as such.

This appears to be a bug with the GridView component. To resolve we may have to upgrade to the latest version of the component.

ramonsmits:
Fixes #882, styling logic resulted in multiple rows to be selected based on `dxg:DataViewBase.IsFocused` which is flawed. Also removed the black styling of the row to make it unified with other table/grid controls.

Resolves this behavior:

Gl7aepMUQs

### DIFF
--- a/src/ServiceInsight/App.xaml
+++ b/src/ServiceInsight/App.xaml
@@ -59,8 +59,6 @@
                         <GradientStop Offset="0" Color="#FFFFFAE9" />
                         <GradientStop Offset="1" Color="#FFFFE8A6" />
                     </LinearGradientBrush>
-                    <SolidColorBrush x:Key="FocusedRowBackgroundBrush" Color="#222222" />
-                    <SolidColorBrush x:Key="FocusedRowForegroundBrush" Color="White" />
                     <SolidColorBrush x:Key="ZoomBoxTicksBrush" Color="#FF000000" />
                     <SolidColorBrush x:Key="ZoomBoxButtonBorderBrush" Color="#FFAAACA5" />
                     <SolidColorBrush x:Key="DialogButtonHoverBrush" Color="#FFBCC7D8" />
--- a/src/ServiceInsight/App.xaml
+++ b/src/ServiceInsight/App.xaml
@@ -59,6 +59,8 @@
                         <GradientStop Offset="0" Color="#FFFFFAE9" />
                         <GradientStop Offset="1" Color="#FFFFE8A6" />
                     </LinearGradientBrush>
+                    <SolidColorBrush x:Key="FocusedRowBackgroundBrush" Color="#222222" />
+                    <SolidColorBrush x:Key="FocusedRowForegroundBrush" Color="White" />
                     <SolidColorBrush x:Key="ZoomBoxTicksBrush" Color="#FF000000" />
                     <SolidColorBrush x:Key="ZoomBoxButtonBorderBrush" Color="#FFAAACA5" />
                     <SolidColorBrush x:Key="DialogButtonHoverBrush" Color="#FFBCC7D8" />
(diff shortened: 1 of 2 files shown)


mauroservienti (approved):

src/ServiceInsight/MessageList/MessageListView.xaml
```diff
@@ -12,20 +12,6 @@
...
              d:DesignWidth="500"
              mc:Ignorable="d">
-    <UserControl.Resources>
-        <Style x:Key="FocusedRowStyle" TargetType="dxg:RowControl">
-            <Style.Triggers>
-                <DataTrigger Binding="{Binding Path=IsSelected}" Value="True">
-                    <Setter Property="Background" Value="{StaticResource FocusedRowBackgroundBrush}" />
-                    <Setter Property="Foreground" Value="{StaticResource FocusedRowForegroundBrush}" />
-                </DataTrigger>
-                <Trigger Property="dxg:DataViewBase.IsFocused" Value="True">
```
WojcikMike, line 22: Only this section needs to be removed. If you remove the whole style, the selection on the grid has the default blue coloring. The brushes, Rowstyle and IsSelected binding should stay

WojcikMike (changes requested):

src/ServiceInsight/App.xaml
```diff
@@ -59,8 +59,6 @@
                         <GradientStop Offset="0" Color="#FFFFFAE9" />
                         <GradientStop Offset="1" Color="#FFFFE8A6" />
                     </LinearGradientBrush>
-                    <SolidColorBrush x:Key="FocusedRowBackgroundBrush" Color="#222222" />
-                    <SolidColorBrush x:Key="FocusedRowForegroundBrush" Color="White" />
```
mauroservienti, line 63: As @WojcikMike correctly points out in the comment below this section needs to stay otherwise the trigger

```
<DataTrigger Binding="{Binding Path=IsSelected}" Value="True">
```

cannot set the expected values.

src/ServiceInsight/MessageList/MessageListView.xaml
mauroservienti: Good catch. The resulting style should be like the following:

```
<Style x:Key="FocusedRowStyle" TargetType="dxg:RowControl">
  <Style.Triggers>
    <DataTrigger Binding="{Binding Path=IsSelected}" Value="True">
      <Setter Property="Background" Value="{StaticResource FocusedRowBackgroundBrush}" />
      <Setter Property="Foreground" Value="{StaticResource FocusedRowForegroundBrush}" />
    </DataTrigger>
  </Style.Triggers>
</Style>
```

I'd also suggest renaming the style to `SelectedRowStyle` and updating line 220 accordingly.

mauroservienti (changes requested):

src/ServiceInsight/MessageList/MessageListView.xaml
ramonsmits: I just restored it to the previous name as otherwise its a name refactoring.

ramonsmits:
Created https://github.com/Particular/ServiceInsight/issues/1224 as the black/white style is very different from the other blue active versus gray deactive styles

PR #1217, opened 2022-06-04
Make ServiceControl 500 errors visible

WojcikMike:


src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
```diff
@@ -707,7 +711,7 @@ void LogResponse(IRestResponse response)
         void LogError(IRestResponse response)
         {
             var exception = response?.ErrorException;
-            var errorMessage = response != null ? string.Format("Error executing the request: {0}, Status code is {1}", response.ErrorMessage, response.StatusCode) : "No response was received.";
+            var errorMessage = response != null ? string.Format("Error executing the request: {0}, Status code is {1}, content: {2}", response.ErrorMessage, response.StatusCode, response.Content) : "No response was received.";
```
mauroservienti, line 714: Not sure it compiles, I wrote it in the browser. This reads better than the string format positional arguments.

```suggestion
            var errorMessage = response != null ? $"Error executing the request: {response.ErrorMessage}, Status code is {response.StatusCode}, content: {response.Content}" : "No response was received.";
```

mauroservienti (approved):

src/ServiceInsight/ServiceControl/DefaultServiceControl.cs
WojcikMike: I changed it in Rider, as I want to make sure it compiles ;)

PR #1218, opened 2022-06-07
Fix flow diagram message item cropping

mauroservienti:
The flow diagram crops the message icon when displaying messages:

Screen Shot 2022-06-21 at 14 21 04


WojcikMike:
Currently the lines connecting the items are starting within the box: 
image

And here:
image

HEskandari:
@WojcikMike To be clear, they were always inside the 'box'. The box did not have a visible border, which is why the impression was that the arrows start outside of the box.

mauroservienti:
This has not been included in the `2.10.0` release. Removing assignees and reviewers.

WojcikMike:
@mauroservienti should we close this PR? otherwise we will close it when stalebot finds it

mauroservienti:
yeah, we can close it. Even if I like to annoy that bot 😄

PR #1221, opened 2022-06-09
Fix search from command line

### LINKED ISSUE #1211: Search results are not displayed when clicking on 'View in ServiceInsight" from ServicePulse
## Steps to reproduce

- Open ServiceInsight
  - Manually connect to a ServiceControl instance
  - Make sure the instance is the selected one
- Open ServicePulse connected to a different ServiceControl instance
  - Select a failed message
  - Click on `View in ServiceInsight`
- ServiceInsight will open (or get to the foreground)
  - Si will add the new connection
  - BUG: the new connection is not selected in the explorer tree view (it's a different bug tracked in #1212 )
  - SI will perform the search
  - the failed message is briefly displayed in the messages view
  - the messages view is refreshed and all messages are listed

From the second time on, the behavior is correct. To trigger the bug again it's enough to change the ServiceControl selected node in the Endpoint explorer and the bug manifests itself once more.

## Expected behavior

ServiceInsight displays the failed message and not all the messages from the recently added ServiceControl connection.

mauroservienti:
This is a dirty trick to fix #1211.

As described in the issue when ServicePulse requires to execute a search ServiceInsight also tries to select a node in the Endpoint explorer. When that happens the messages list view reacts and tries to load messages for the selected node. At the time, ServiceInsight performs a search based on ServicePulse arguments. The search is faster than the messages refresh; thus, the messages list gets replaced by the operation that finishes last, which is always performed by the messages list view overwriting the search results.

### DIFF
--- a/src/ServiceInsight/MessageList/MessageListViewModel.cs
+++ b/src/ServiceInsight/MessageList/MessageListViewModel.cs
@@ -194,6 +194,11 @@
             selectedExplorerItem = @event.SelectedExplorerItem;
             ServiceControl = @event.SelectedExplorerItem.GetServiceControlClient(clientRegistry);
 
+            if (Parent.SearchInProgress)
+            {
+                return;
+            }
+
             if (ServiceControl != null)
             {
                 await RefreshMessages();
--- a/src/ServiceInsight/Shell/ShellViewModel.cs
+++ b/src/ServiceInsight/Shell/ShellViewModel.cs
@@ -477,10 +477,19 @@
             DisplayRegistrationStatus();
         }
 
-        public async Task PostConfigurationUpdate()
+        public async Task PerformSearchAndInitializeRefreshTimer()
         {
-            await Messages.SearchBar.PerformCommandLineSearch();
+            try
+            {
+                await Messages.SearchBar.PerformCommandLineSearch();
+            }
+            finally
+            {
+                SearchInProgress = false;
+            }
             InitializeAutoRefreshTimer();
         }
+
+        public bool SearchInProgress { get; set; }
     }
 }
--- a/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
+++ b/src/ServiceInsight/Explorer/EndpointExplorer/EndpointExplorerViewModel.cs
@@ -291,8 +291,9 @@
 
         public async Task Handle(ConfigurationUpdated message)
         {
+            Parent.SearchInProgress = true;
             await ConnectToService(commandLineParser.ParsedOptions.EndpointUri.ToString());
-            await Parent.PostConfigurationUpdate();
+            await Parent.PerformSearchAndInitializeRefreshTimer();
         }
     }
 }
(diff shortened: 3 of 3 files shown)


src/ServiceInsight/MessageList/MessageListViewModel.cs
```diff
@@ -194,6 +194,11 @@ public async Task Handle(SelectedExplorerItemChanged @event)
             selectedExplorerItem = @event.SelectedExplorerItem;
             ServiceControl = @event.SelectedExplorerItem.GetServiceControlClient(clientRegistry);
 
+            if (Parent.SearchInProgress)
+            {
+                return;
+            }
+
```
ramonsmits, line 201: I do not see `SearchInProgress` set to `true` anywhere. Should it be here?

```suggestion
            Parent.SearchInProgress = true;
```

src/ServiceInsight/Shell/ShellViewModel.cs
```diff
@@ -477,10 +477,19 @@ public virtual void Handle(LicenseUpdated message)
             DisplayRegistrationStatus();
         }
 
-        public async Task PostConfigurationUpdate()
+        public async Task PerformSearchAndInitializeRefreshTimer()
         {
-            await Messages.SearchBar.PerformCommandLineSearch();
+            try
+            {
+                await Messages.SearchBar.PerformCommandLineSearch();
```
ramonsmits, line 484: or here?

```suggestion
                SearchInProgress = true;
                await Messages.SearchBar.PerformCommandLineSearch();
```

ramonsmits (changes requested):

mauroservienti:
@ramonsmits that's horrible, I know. I discussed the options with Hadi last week, and he suggested using the `MessageSelectionContext` to keep track of the search status in a single place. We just pushed https://github.com/Particular/ServiceInsight/pull/1221/commits/686cdbf29a0225a66a282e087be71c56d5f9417c that does that

src/ServiceInsight/Search/SearchBarViewModel.cs
```diff
@@ -202,7 +203,7 @@ public async Task RefreshResult()
 
         public int TotalPagesCount { get; private set; }
 
-        public bool SearchInProgress { get; private set; }
+        //public bool SearchInProgress { get; private set; }
```
ramonsmits, line 207:
```suggestion
```

src/ServiceInsight/MessageList/MessageSelectionContext.cs
```diff
@@ -15,6 +15,8 @@ public MessageSelectionContext(IEventAggregator eventAggregator)
 
         public StoredMessage SelectedMessage { get; set; }
 
+        public bool SearchInProgress { get; set; }
```
ramonsmits, line 18:
```suggestion
        public bool SearchInProgress { get; set; } // Serializes UI search triggers that prevents UI message re-loading
```

ramonsmits (approved):
Minor suggestion to improve the in code comments to explain why it exists.

WojcikMike (approved):

PR #1448, opened 2024-04-25
Add NuGetAuditMode = all and bump necessary versions

tamararivera:
Part of the work for https://github.com/Particular/PlatformInternals/issues/604


DavidBoike:
No release needed since the only update was in tests.

PR #1572, opened 2025-03-14
Add ability to use non-default username and password for Windows Auth

Beau-Gosse-LettrLabs:
Allows users who follow the directions to [Install ServicePulse in IIS](https://docs.particular.net/servicepulse/install-servicepulse-in-iis) to allow users to provide a username and password other than the one they're currently logged in as locally.

 Screenshot 2025-03-13 193018

### DIFF
--- a/src/ServiceInsight.Tests/ServiceControlConnectionDialogTests.cs
+++ b/src/ServiceInsight.Tests/ServiceControlConnectionDialogTests.cs
@@ -1,5 +1,6 @@
 ﻿namespace ServiceInsight.Tests
 {
+    using System.Security;
     using System.Threading.Tasks;
     using Autofac;
     using Caliburn.Micro;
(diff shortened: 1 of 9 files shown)


Beau-Gosse-dev:
@DavidBoike Is there anything I can change or update on this PR to get it merged? We really need this feature

DavidBoike:
@Beau-Gosse-dev did you see that we released ServicePulse 2.0 which includes (as far as I'm aware) all of the functionality present in ServiceInsight in ServicePulse? Then this issue goes away because the browser takes care of all the authentication concerns.

https://discuss.particular.net/t/servicepulse-2-0-0-major-release-available/4327

Now that that's available, I'm not sure if there will be a lot of reasons to accept updates to the ServiceInsight code like this.

PR #1679, opened 2026-07-13
Remove the filter by actor from the release workflow

tamararivera:
This PR removes the filter by actor, given that was added because Dependabot wasn't allow to sign the artifacts. Now Renovate can handle that, so we don't need the filter anymore.


tamararivera:
Closing as it is superseded by #1680

PR #1691, opened 2026-07-23
Remove Setup project from solution

bording:
This PR moves the Setup project from being in the solution as a project that is configured not to build and instead includes it as a Solution item file. 

This lets you still see the file to make changes to it, and should also remove it from consideration at all went interacting at the solution-level, which should fix the NuGet audit workflow.


bording:
I'm going to merge this in even with the package failures in order to test the NuGet audit workflow with this change.

